PR #1, opened 2022-05-11
Add Prettier formatting

tsa96:
Adds a prettier config file and sets up Husky and lint-staged to run after each commit.

Adds Prettier scripts to package.json including a check for if we ever want to integrate it with Github Actions (probably overkill, up to you Goc).

Reformats all our JS, XML, SCSS and Markdown with Prettier. 

God it feels good to be back to Github PRs.


Gocnak:
Let's add the github action to check lint output in this PR.

tsa96:
Added the workflow for the action, and tested it using [Act](https://github.com/nektos/act)

.github/workflows/format.yml
```diff
@@ -0,0 +1,20 @@
+name: Formatting
+
+on:
+  push:
+    branches: [ master ]
```
Gocnak, line 5: Doesn't need to happen on push to master, only on PRs to master, as we're doing trunk-based development.

.github/workflows/format.yml
```diff
@@ -0,0 +1,20 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Use Node.js 18
+      uses: actions/setup-node@v3
+      with:
+        node-version: '18'
+        cache: 'npm'
+    - run: npm ci
```
Gocnak, line 19: What does `npm ci` do?

Gocnak (changes requested):

.github/workflows/format.yml
```diff
@@ -0,0 +1,20 @@
...
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Use Node.js 18
```
Gocnak, line 14: Nit
```suggestion
    - name: Setup Node.js 18

```

.github/workflows/format.yml
tsa96: Was getting a bit confused over if I should be running `npm ci` or `npm install`. Been reading up on it and I think the fastest setup for us is first caching the `node_modules/` then running `npm install`, which grabs the dependencies if they're not cached.

.github/workflows/format.yml
tsa96: Done

.github/workflows/format.yml
tsa96: Done

Gocnak (approved):

PR #10, opened 2022-05-28
fix: Move settings tabs to common file

tsa96:
Both the main settings and the search js files use this to I moved it to a separate file. Fixes a JS error that occurs when trying to search settings.


layout/mainmenu_settings.xml
```diff
@@ -4,6 +4,7 @@
 	</styles>
 
 	<scripts>
+		<include src="file://{resources}/scripts/common/settings.js" />
 		<include src="file://{resources}/scripts/settings_shared.js" />
```
Gocnak, line 8: Should settings_shared stuff be moved to common/settings.js ?

tsa96: Mmh no, settings_shared is actually a bunch of unique functionality for different settings pages, the structure is a carryover from CS:GO's settings. I'd quite like to refactor that but too much work for little gain currently IMO.

scripts/common/settings.js
```diff
@@ -0,0 +1,68 @@
+const SETTINGS_TABS = {
```
Gocnak, line 1: Missing `'use strict';`
```suggestion
'use strict';

const SETTINGS_TABS = {
```

Gocnak (changes requested):

scripts/common/settings.js
tsa96: Fixed

Gocnak (approved):

PR #15, opened 2022-06-14
Add background image for triple monitor mode

SlidyBat:


Gocnak (approved):
Old branding but we can always adjust all our textures like this in the future towards 1.0.0.

tsa96 (commented):
Our other background images are DDS, I assume Panorama converts them back wouldn't it be best to do the same here?

Gocnak:
PNG should be fine, the platform code calls for it, and it's already really small at 300 kb. The bigger background textures that are more in-focus can be DDS for the compression savings.

PR #21, opened 2022-06-28
feat: better cgaz

PeenScreeker:
** C++ CHANGE REQUIRED **
This change:
- improves cgaz responsiveness
- cleans up ConVarEnabler use
- adds a setting for coloring the "uncolored" snap zones
- fixes vector normalization bug
- and sets the max height multiplier to 1x setting for sizing snaps by gain

### DIFF
--- a/layout/hud/cgaz.xml
+++ b/layout/hud/cgaz.xml
@@ -49,7 +49,12 @@
     </Panel>
 		<Panel id="WindicatorArrow">
       <ConVarEnabler class="full-height" convar="mom_df_hud_windicator_enable" togglevisibility="true">
-        <Image id="WindicatorArrowIcon" class="arrow__down" src="file://{images}/hud/cgaz-arrow.svg" textureheight="48" />
+        <Image
+					id="WindicatorArrowIcon"
+					class="arrow__down"
+					src="file://{images}/hud/cgaz-arrow.svg"
+					textureheight="48"
+				/>
       </ConVarEnabler>
     </Panel>
 	</MomHudCgaz>
--- a/layout/hud/cgaz.xml
+++ b/layout/hud/cgaz.xml
@@ -8,7 +8,7 @@
 		<include src="file://{resources}/scripts/hud/cgaz.js" />
 	</scripts>
 
-	<MomHudCgaz class="full-width full-height">
+	<MomHudCgaz class="full-width full-height" alternateTicks="false">
 		<Panel id="AccelContainer" class="cgaz__container">
 			<ConVarEnabler class="cgaz" convar="mom_df_hud_accel_enable" togglevisibility="true">
 				<Panel id="LeftTurnZone" />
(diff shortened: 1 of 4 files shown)


PeenScreeker:
Implements some features from open issue #1773.

scripts/hud/cgaz.js
```diff
@@ -88,17 +88,7 @@ class Cgaz {
...
-				'ChaosHudProcessInput',
-				$.GetContextPanel(),
-				this.onUpdate.bind(this)
-			);
-		} else {
-			this.velocityArrow.visible = false;
-			this.windicatorArrow.visible = false;
-			return;
-		}
+		if (GAMEMODE_WITH_NULL[GameModeAPI.GetCurrentGameMode()].name !== 'Defrag') return;
```
Gocnak, line 91: This is really weird. We should define an enum object that follows the C++ style of gamemodes, so you can do `GameModeAPI.GetCurrentGameMode() !== GameMode.DEFRAG` instead. This can be done in another PR, so this is just a nit for now.

Gocnak (approved):

scripts/hud/cgaz.js
Gocnak: Created an issue for it at #23

scripts/hud/cgaz.js
```diff
@@ -718,17 +726,19 @@ class Cgaz {
 
 	static getNormal(vec, threshold) {
 		const mag = this.getSize(vec);
+		let vecNormal = {
```
tsa96, line 729: This can be `const`, the properties in a JS object are still mutable. Weird I know, but good practice

layout/hud/cgaz.xml
```diff
@@ -8,9 +8,9 @@
 		<include src="file://{resources}/scripts/hud/cgaz.js" />
 	</scripts>
 
-	<MomHudCgaz class="full-width full-height">
+	<MomHudCgaz class="full-width full-height" alternateTicks="false">
 		<Panel id="AccelContainer" class="cgaz__container">
-			<ConVarEnabler class="cgaz" convar="mom_df_hud_accel_enable" togglevisibility="true">
```
tsa96, line 13: If we're removing this class we should probably rethink `cgaz.scss` a bit. We could only have the `cgaz__container` class, but I'm inclined to say we just remove it and use `full-width` and `horizontal-align-left` instead of the `cgaz__container` class.

tsa96 (changes requested):

layout/hud/cgaz.xml
PeenScreeker: Ok, can we make an issue for this and review with the rest of the upcoming BetterCgaz(tm) changes?

PeenScreeker: #31 captures this comment

tsa96 (approved):

PR #34, opened 2022-09-02
feat: Add Jump Sync defrag hud element

PeenScreeker:
This pull request adds a HUD element to aid jump timing in defrag.

- `mom_hud_jump_sync_enable` to draw jump sync bars - upward for late release, downward for early/late press.

- `mom_hud_jump_sync_max_delay` to set the max value for tracking jump presses. This is the value corresponding to 100% on either bar.

- `mom_hud_jump_sync_text_enable` to show time (ms) next to bars.

Requires Red changes to merge. Closes [#1793](https://github.com/momentum-mod/game/issues/1793)

### DIFF
--- a/scripts/hud/dfjump.js
+++ b/scripts/hud/dfjump.js
@@ -1,0 +1,59 @@
+'use strict';
+
+const COLOR_CLASS = {
+	AIR: 'dfjump__press--air',
+	GROUND: 'dfjump__press--ground'
+};
+
+const DEFAULT_DELAY = 360;
+
+class DFJump {
+	/** @type {Panel} @static */
+	static container = $('#DFJumpContainer');
+	/** @type {ProgressBar} @static */
+	static releaseBar = $('#JumpReleaseBar');
+	/** @type {ProgressBar} @static */
+	static pressBar = $('#JumpPressBar');
+	/** @type {Label} @static */
+	static releaseLabel = $('#JumpReleaseLabel');
+	/** @type {Label} @static */
+	static pressLabel = $('#JumpPressLabel');
+	/** @type {Label} @static */
+	static totalLabel = $('#JumpTotalLabel');
+
+	static onLoad() {
+		this.initializeSettings();
+		this.colorClass = COLOR_CLASS.GROUND;
+	}
+
+	static onDFJumpUpdate(releaseDelay, pressDelay, totalDelay) {
+		const releaseRatio = releaseDelay * this.inverseMaxDelay;
+		const pressRatio = Math.abs(pressDelay) * this.inverseMaxDelay;
+		const newPressColorClass = pressDelay < 0 ? COLOR_CLASS.GROUND : COLOR_CLASS.AIR;
+
+		this.releaseBar.value = releaseRatio;
+		this.pressBar.value = pressRatio;
+		this.pressBar.RemoveClass(this.colorClass);
+		this.pressBar.AddClass(newPressColorClass);
+		this.colorClass = newPressColorClass;
+
+		this.releaseLabel.text = releaseDelay.toFixed();
+		this.pressLabel.text = pressDelay.toFixed();
+		this.totalLabel.text = totalDelay.toFixed();
+	}
+
+	static setMaxDelay(newDelay) {
+		this.inverseMaxDelay = 1.0 / (newDelay ?? DEFAULT_DELAY);
+	}
+
+	static initializeSettings() {
+		this.setMaxDelay(GameInterfaceAPI.GetSettingInt('mom_df_hud_jump_max_delay'));
+	}
+
+	static {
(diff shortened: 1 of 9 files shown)


scripts/hud/dfjump.js
```diff
@@ -0,0 +1,85 @@
...
+	ATTACK2: 1 << 11,
+	SCORE: 1 << 16,
+	SPEED: 1 << 17,
+	WALK: 1 << 18,
+	ZOOM: 1 << 19,
+	LOOKSPIN: 1 << 25,
+	BHOPDISABLED: 1 << 29,
+	PAINT: 1 << 30,
+	STRAFE: 1 << 31
+};
```
braem, line 24: This def should be in a common file like scripts/common/buttons.js

scripts/hud/dfjump.js
```diff
@@ -0,0 +1,85 @@
...
+		this.pressBar.AddClass(newPressColorClass);
+		this.colorClass = newPressColorClass;
+
+		this.releaseLabel.text = releaseDelay.toFixed();
+		this.pressLabel.text = pressDelay.toFixed();
+		this.totalLabel.text = totalDelay.toFixed();
+	}
+
+	static setMaxDelay(newDelay) {
+		this.maxDelay = this.nullCheck(newDelay, 1);
```
braem, line 67: How can the delay be null?

styles/config.scss
```diff
@@ -661,6 +661,19 @@ $speedometer-event-fadeout-time: 5s !default;
...
+// ================
+$dfjump-width: 20px;
+$dfjump-height: 300px;
+$dfjump-opacity: 0.7;
+$dfjump-color-release: gradient(linear, 30% 0%, 100% 0%, from(#ff5757), to(#d31818)) !default;
+$dfjump-color-press-air: gradient(linear, 30% 0%, 100% 0%, from(#1896d3), to(#3f4aca)) !default;
+$dfjump-color-press-ground: gradient(linear, 30% 0%, 100% 0%, from(#7aee7a), to(#159856)) !default;
+$dfjump-color-background: gradient(linear, 0% 0%, 100% 0%, from(#5f5f5f80), to(#c8c8c880)) !default;
+$dfjump-font-size: 22px;
+$dfjump-padding: 6px;
```
braem, line 675: Default settings should all have `!default`

styles/hud/dfjump.scss
```diff
@@ -0,0 +1,82 @@
...
+			vertical-align: bottom;
+		}
+
+		&--total {
+			vertical-align: middle;
+		}
+	}
+}
+
+#DFJumpContainer {
```
braem, line 78: I think we're still following a strict no styling on IDs? Not sure. @tsa96 halp

braem (changes requested):

styles/hud/dfjump.scss
PeenScreeker: no sweat. I've changed this to a class applied in the xml file

scripts/hud/dfjump.js
PeenScreeker: Idk, I didn't really dig into how the console sanitizes input. Just traumatized from all the JS errors in cgaz.js from something trying to set a panel value to `NaN px` - it makes a difference there. Since this is the denominator, I wanted to guarantee there wouldn't ever be funny business, Not a big deal, I could remove this here if it's really not needed.

braem: GetInt on a cvar will thankfully always return an int. Might wanna check if it's 0 though? Unless the cvar has appropriate bounds. Or might be better to check if `inverseMaxDelay` is nan

scripts/hud/dfjump.js
```diff
@@ -0,0 +1,62 @@
...
+		this.inverseMaxDelay = 1.0 / this.maxDelay;
+	}
+
+	static nullCheck(val, def) {
+		return val === null ? def : val;
+	}
```
Gocnak, line 50: Welcome to the world of JavaScript, where you can simply do `return val ?? def` here, the "nullish coalescing" operator: https://sebhastian.com/javascript-double-question-mark/

In fact, the function basically becomes unnecessary then and all invokes can be replaced with the double question mark.

You will want to test this, I believe our V8 version is new enough to handle it.

Gocnak (changes requested):

scripts/hud/dfjump.js
PeenScreeker: wow. This is really nice for this use case.

I'll remove the `nullCheck()` method.

scripts/hud/dfjump.js
PeenScreeker: the minimum value allowed for the max delay convar is 1. Goc's suggestion to use the ?? operator removes the need for this function, so I'll get rid of it.

I've updated the line to use a better default instead of a magic number.

Gocnak (approved):

braem (approved):

PR #35, opened 2022-09-06
feat: Use haste timer to get snapzone shapes

PeenScreeker:
This PR fixes an old hack where snap shapes were only calculated at the first player input. Snap shapes are now correctly drawn when the Haste powerup is used.

Requires changes on Red.

### DIFF
--- a/scripts/hud/cgaz.js
+++ b/scripts/hud/cgaz.js
@@ -3,6 +3,8 @@
 // TODO: remove these globals
 let MAX_GROUND_SPEED = 320;
 const AIR_ACCEL = 1;
+const DEFAULT_ACCEL = 2.56;
+const HASTE_ACCEL = 3.328; // (max speed) * (air accel = 1) * (tick interval) * (haste factor)
 
 let NEUTRAL_CLASS;
 let SLOW_CLASS;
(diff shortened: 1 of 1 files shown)


scripts/hud/cgaz.js
```diff
@@ -250,6 +250,12 @@ class Cgaz {
 		const accel = lastMoveData.acceleration;
 		const maxAccel = accel * maxSpeed * tickInterval;
 
+		if (lastMoveData.hasteEndTime < 0 || lastMoveData.hasteEndTime > MomentumMovementAPI.GetCurrentTime()) {
+			this.snapAccel = 3.328; // (max speed) * (air accel = 1) * (tick interval) * (haste factor)
+		} else {
+			this.snapAccel = 2.56;
```
chovelyoukai, line 256: Is there a way to change these to constants instead of magic numbers?

PeenScreeker: yep, good look. I've defined these as constants now

braem (approved):

chovelyoukai (approved):

Gocnak (approved):

PR #38, opened 2022-10-03
Draft: Defrag project drawdelay setting

tsa96:
Leaving as a draft til Lumia confirms if there's any other defrag weapon stuff we could add here.

Once in, add to POEditor
```
"Settings_Defrag_Projectile_DrawDelay": "Projectile render delay",
"Settings_Defrag_Projectile_DrawDelay_Info": "The delay in seconds before all defrag projectiles are rendered on-screen."
```

### DIFF
--- a/layout/settings/settings_gameplay.xml
+++ b/layout/settings/settings_gameplay.xml
@@ -468,7 +468,7 @@
 					text="#Settings_RJ_RocketDrawDelayEntry"
 					min="0"
 					max="1"
-					displayprecision="1"
+					displayprecision="2"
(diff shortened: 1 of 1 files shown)


tsa96:
Maybe we don't want drawdelay actually cus don't spawn inside head like TF2

tsa96:
Okay we're avoiding the defrag draw delay since not as important as TF2 mode and conc doesn't use either, just updating my branch so Mac can work from it

tsa96:
Being done in Mac's PR

PR #39, opened 2022-10-09
Pano showfps, showpos, and showbonecounts panels

braem:
Closes https://github.com/momentum-mod/game/issues/1516

Merge with red PR

### DIFF
--- a/styles/components/showfps.scss
+++ b/styles/components/showfps.scss
@@ -1,0 +1,34 @@
+@use '../config' as *;
+@use '../abstract/mixin';
+
+.fpsmeter {
+	width: 100%;
+	height: 100%;
+	align: left top;
+
+	&__label {
+		@include mixin.font-styles($use-header: true);
+		font-size: 16px;
+		text-align: left;
+		text-overflow: ellipsis;
+		color: $white;
+		text-shadow-fast: $font-shadow;
+		font-family: $font-monospace;
+
+		&.ExcellentFPS {
+			color: $green;
+		}
+		&.GoodFPS {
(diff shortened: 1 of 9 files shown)


styles/components/showfps.scss
```diff
@@ -0,0 +1,34 @@
+@use '../config' as *;
+@use '../abstract/mixin';
+
+.fpsmeter {
+	width: 100%;
```
tsa96, line 5: Does this need to be `width: 100%`/`height: 100%`? Even if it's top-level can't we just position without that?

tsa96 (changes requested):

styles/components/showfps.scss
```diff
@@ -0,0 +1,34 @@
...
+	height: 100%;
+	align: left top;
+
+	&__label {
+		@include mixin.font-styles($use-header: true);
+		font-size: 16px;
+		text-align: left;
+		text-overflow: ellipsis;
```
PeenScreeker, line 13: not sure if it matters, `ellipsis` is the default mode for `text-overflow`.

PeenScreeker (approved):

styles/components/showfps.scss
braem: We can yeah, if we ever want other panels below it. I'd like it to be absolutely positioned in a place that makes more sense for UI in the future

Gocnak: Yeah I think this is fine until we do more absolute positioning (HUD customization)

Gocnak (approved):

PR #40, opened 2022-10-17
feat: jump stats

PeenScreeker:
This PR introduces a panel to implement jump stats like the SSJ plugin popular in CS:S. The ability to show jump stats on the strafe synchronizer is also introduced in this PR.

Jump stat settings:
- `mom_hud_jstats_enable` display bhop jump stats
- `mom_hud_jstats_first` first jump to log stats on jump stats panel
- `mom_hud_jstats_interval` log stats every X jumps after first log
- `mom_hud_jstats_takeoff_speed_enable` show takeoff speed on jump stats panel
- `mom_hud_jstats_speed_delta_enable` show speed delta on jump stats panel
- `mom_hud_jstats_takeoff_time_enable` show takeoff time on jump stats panel
- `mom_hud_jstats_time_delta_enable` show time delta on jump stats panel
- `mom_hud_jstats_strafe_sync_enable` show sync on jump stats panel
- `mom_hud_jstats_strafe_count_enable` show strafe count on jump stats panel
- `mom_hud_jstats_speed_gain_enable` show speed gain on jump stats panel
- `mom_hud_jstats_strafe_ratio_enable` show strafe ratio on jump stats panel
- `mom_hud_jstats_height_delta_enable` show height delta on jump stats panel
- `mom_hud_jstats_efficiency_enable` show jump efficiency on jump stats panel

New synchronizer settings:
- `mom_hud_synchro_stat_mode` show jump stats on strafe synchronizer. 0 = OFF, 1 = Show stats, 2 = Hide bar
- `mom_hud_synchro_dynamic_enable` enable bar direction to change with movement inputs (default ON). Turning this setting off causes the bar to fill from one side no matter the current input direction

Requires red and green changes.

### DIFF
--- a/layout/hud/jumpstats.xml
+++ b/layout/hud/jumpstats.xml
@@ -83,13 +83,13 @@
 				<Label text="{s:gain}" id="GainLabel" class="jumpstats__label--values" />
 			</ConVarEnabler>
 			<ConVarEnabler
-				id="StrafeRatioContainer"
-				class="jumpstats__label jumpstats__label--strafe-ratio"
-				convar="mom_hud_jstats_strafe_ratio_enable"
+				id="YawRatioContainer"
+				class="jumpstats__label jumpstats__label--yaw-ratio"
+				convar="mom_hud_jstats_yaw_ratio_enable"
 				togglevisibility="true"
 			>
-				<Label text="#JumpStats_Label_StrafeRatio" class="jumpstats__label--name" />
-				<Label text="{s:strafe_ratio}" id="StrafeRatioLabel" class="jumpstats__label--values" />
+				<Label text="#JumpStats_Label_YawRatio" class="jumpstats__label--name" />
+				<Label text="{s:yaw_ratio}" id="YawRatioLabel" class="jumpstats__label--values" />
 			</ConVarEnabler>
 			<ConVarEnabler
 				id="HeightDeltaContainer"
--- a/layout/hud/jumpstats.xml
+++ b/layout/hud/jumpstats.xml
@@ -25,7 +25,7 @@
 				convar="mom_hud_jstats_takeoff_speed_enable"
 				togglevisibility="true"
 			>
-				<Label text="Speed" class="jumpstats__label--name" />
+				<Label text="#JumpStats_Label_Speed" class="jumpstats__label--name" />
 				<Label text="{s:speed}" id="SpeedLabel" class="jumpstats__label--values" />
 			</ConVarEnabler>
 			<ConVarEnabler
@@ -34,7 +34,7 @@
 				convar="mom_hud_jstats_speed_delta_enable"
 				togglevisibility="true"
 			>
-				<Label text="ΔS" class="jumpstats__label--name" />
+				<Label text="#JumpStats_Label_DeltaSpeed" class="jumpstats__label--name" />
 				<Label text="{s:speed_delta}" id="SpeedDeltaLabel" class="jumpstats__label--values" />
 			</ConVarEnabler>
 			<ConVarEnabler
@@ -43,7 +43,7 @@
 				convar="mom_hud_jstats_takeoff_time_enable"
 				togglevisibility="true"
 			>
-				<Label text="Time" class="jumpstats__label--name" />
+				<Label text="#JumpStats_Label_Time" class="jumpstats__label--name" />
 				<Label text="{s:time}" id="TimeLabel" class="jumpstats__label--values" />
 			</ConVarEnabler>
 			<ConVarEnabler
@@ -52,7 +52,7 @@
 				convar="mom_hud_jstats_time_delta_enable"
 				togglevisibility="true"
 			>
-				<Label text="ΔT" class="jumpstats__label--name" />
+				<Label text="JumpStats_Label_DeltaTime" class="jumpstats__label--name" />
 				<Label text="{s:time_delta}" id="TimeDeltaLabel" class="jumpstats__label--values" />
 			</ConVarEnabler>
 			<ConVarEnabler
@@ -61,7 +61,7 @@
 				convar="mom_hud_jstats_strafe_count_enable"
 				togglevisibility="true"
 			>
-				<Label text="Strafes" class="jumpstats__label--name" />
+				<Label text="#JumpStats_Label_Strafes" class="jumpstats__label--name" />
 				<Label text="{s:strafes}" id="StrafesLabel" class="jumpstats__label--values" />
 			</ConVarEnabler>
 			<ConVarEnabler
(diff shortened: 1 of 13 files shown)


scripts/hud/jumpstats.js
```diff
@@ -0,0 +1,133 @@
...
+			this.strafeRatioBuffer.reduce((sum, element) => sum + element)
+		);
+		this.panel.SetDialogVariable(
+			'height_delta',
+			this.heightDeltaBuffer.reduce((sum, element) => sum + element)
+		);
+		this.panel.SetDialogVariable(
+			'efficiency',
+			this.efficiencyBuffer.reduce((sum, element) => sum + element)
+		);
```
tsa96, line 117: Worth optimising this a little, since it runs so frequently. Every time you pass an arrow function to `reduce` V8 is going to create a new function (I don't know exactly how this is represented in memory, there's probably some crazy optimisations, but let's clear this up anyway). So we can declare a separate method
```js
sumBuffer(buffer) {
    return buffer.reduce((sum, element) => sum + element);
}
```
FWIW `reduce`s are often slower than a regular imperative for loop, and easier to read. But I think with a super simple case like this it's fine and has the same time complexity.

scripts/mainmenu_mapselector.js
```diff
@@ -134,6 +135,8 @@ class MapSelection {
 		// Set unchecked counter to -1, if you're using this you've got the hang of it, no more popups for you!
 		this.timesModeButtonsUnchecked = -1;
 
+		$.Msg(selectedMode);
```
tsa96, line 138: This was a mistake of mine that's now fixed on master, so you must be rebased wrong

tsa96 (changes requested):
Not sure if this is ready for review but figured I'd get on with in anyway

Also, you need to localize all the English strings here! I can explain how over Discord if needed

scripts/hud/jumpstats.js
```diff
@@ -0,0 +1,133 @@
...
+
+		this.setText();
+	}
+
+	static initializeBuffer(size) {
+		return Array(size).fill('');
+	}
+
+	static addToBuffer(buffer, value) {
+		buffer[buffer.length - 1] += '\n';
```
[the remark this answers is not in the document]
epkieffe: I think `Array.at()` only works to fetch the value in the array at the desired position. Since the purpose here is to assign a value to the last position in the array, something else will be needed. I think python supports negative indices for accessing from the end of an array, but js seems to only allow `array[array.length - 1]`.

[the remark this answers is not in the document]
epkieffe: image

layout/hud/jumpstats.xml
```diff
@@ -0,0 +1,114 @@
...
+				id="CountLabel"
+				class="jumpstats__label jumpstats__label--count jumpstats__label--values"
+			/>
+			<ConVarEnabler
+				id="SpeedContainer"
+				class="jumpstats__label jumpstats__label--speed"
+				convar="mom_hud_jstats_takeoff_speed_enable"
+				togglevisibility="true"
+			>
+				<Label text="Speed" id="" class="jumpstats__label--name" />
```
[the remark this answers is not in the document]
epkieffe: good catch

scripts/hud/jumpstats.js
epkieffe: haha, that's how this was originally implemented. This is how things are done in the synchronizer code, I've implemented this function in jump stats now. Good shout

scripts/mainmenu_mapselector.js
epkieffe: lol found the root cause. I was somehow carrying around your last commit from [refactor: Improve Gamemodes structure](https://github.com/momentum-mod/panorama/pull/40/commits/03db88f29a10540cd303b1fa86c376c3fddf8b7c). Fixed

scripts/hud/jumpstats.js
```diff
@@ -0,0 +1,133 @@
...
+			this.heightDeltaBuffer.reduce((sum, element) => sum + element)
+		);
+		this.panel.SetDialogVariable(
+			'efficiency',
+			this.efficiencyBuffer.reduce((sum, element) => sum + element)
+		);
+	}
+
+	static makeTime(value) {
+		const hours = String((value / 3600).toFixed()).padStart(2, '0');
```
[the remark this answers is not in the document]
epkieffe: there was something broken that seemed to work by using `String()`, but it's obviously not needed anymore. Good look

scripts/hud/jumpstats.js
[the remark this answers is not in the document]
tsa96: Oh my bad, you're completely right!

tsa96:
Okay all specific comments are resolved, but we still need to localize everything. So wherever there's English text (including stuff set from JS) you need to replace with appropriately named tokens (hopefully the naming scheme is obvious from the terms on POEditor), then to import into POEditor, create a json file with structure (there's a copy of this template on TS)

```json
[
    {
        "term": "",
        "definition": ""
    },
    {
        "term": "",
        "definition": ""
    }
]
```

where terms are the token names *not* including `#` at the start, and definitions are the English strings you want. This in POEditor, to go Import, then upload the JSON file, with the option to import definitions set to English.

tsa96 (approved):

layout/mainmenu_settings.xml
```diff
@@ -388,6 +388,15 @@
...
 									<Label class="settings-nav__subitem-label" text="#Settings_WeaponSelection_Title" />
 								</RadioButton>
+								
+								<RadioButton
+									id="JumpStatsRadio"
+									class="settings-nav__subitem"
+									group="SettingsSubNav"
+									onactivate="MainMenuSettings.navigateToSubsection('InterfaceSettings', 'JumpStatsSubSection')"
+								>
+									<Label class="settings-nav__subitem-label" text="Jump Stats" />
```
Gocnak, line 398: Unlocalized?

layout/settings/settings_interface.xml
```diff
@@ -618,7 +618,173 @@
...
+						<RadioButton group="takeoffspeed" text="On" value="1" />
+					</ChaosSettingsEnum>
+
+					<ChaosSettingsEnum
+						text="#Settings_JumpStats_SpeedDelta"
+						convar="mom_hud_jstats_speed_delta_enable"
+						infomessage="#Settings_JumpStats_SpeedDelta_Info"
+					>
+						<RadioButton group="deltaspeed" text="Off" value="0" />
+						<RadioButton group="deltaspeed" text="On" value="1" />
```
Gocnak, line 699: Unlocalized?

Gocnak (changes requested):
Need to localize some strings. Some `On` and `Off`s are already localized (`#Common_On` iirc), but just search for all the text= stuff and make sure everything's localized where it makes sense.

layout/mainmenu_settings.xml
Gocnak: Should probably be `#Settings_JumpStats_Title` right?

PeenScreeker: yep, good look. Missed this file completely when I went through the localization stuff

Gocnak (approved):

PR #42, opened 2022-11-01
Feat: defrag compass (pitch & yaw)

PeenScreeker:
This PR adds settings and functionality for compass hud elements in defrag:

- compass mode - off, show arrow, show ticks, show both
- compass size - compass tick length and arrow half-height
- pitch enable - show/hide pitch target
- pitch target - pitch line offset in degrees (71.8 default for plasma climb)
- compass stat mode - off, show pitch, show yaw, show both
- compass color - normal color for compass elements
- compass highlight color - pitch line & text set to this color when aiming at the pitch target, compass & yaw stat set to this color when velocity is along Pi/4 directions

Requires changes on red to function (cvars and events)

### DIFF
--- a/scripts/hud/cgaz.js
+++ b/scripts/hud/cgaz.js
@@ -12,6 +12,7 @@ let FAST_CLASS;
 let TURN_CLASS;
 let WIN_ZONE_CLASS;
 let MIRROR_CLASS;
+let COMPASS_CLASS;
 
 let COLORED_SNAP_CLASS;
 let UNCOLORED_SNAP_CLASS;
@@ -71,8 +72,15 @@ class Cgaz {
 		$('#SnapZone11')
 	];
 
-	static velocityArrow = $('#VelocityArrow');
-	static velocityArrowIcon = $('#VelocityArrowIcon');
+	static compassArrow = $('#CompassArrow');
+	static compassArrowIcon = $('#CompassArrowIcon');
+	static tickContainer = $('#CompassTicks');
+	static compassTickFull = $('#FullTick');
+	static compassTickHalf = $('#HalfTick');
+	static pitchLine = $('#PitchLine');
+	static pitchStat = $('#PitchStat');
+	static yawStat = $('#YawStat');
+
 	static windicatorArrow = $('#WindicatorArrow');
 	static windicatorArrowIcon = $('#WindicatorArrowIcon');
 	static windicatorZone = $('#WindicatorZone');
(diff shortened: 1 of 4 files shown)


styles/hud/cgaz.scss
```diff
@@ -18,3 +19,54 @@
 		transform: rotatez(180deg);
 	}
 }
+
+.line {
```
tsa96, line 23: Remember, classes defined in Sass files are all compiled into one `main.scss` file. So these class names feel too general, best to use `cgaz-line` etc

styles/hud/cgaz.scss
```diff
@@ -18,3 +19,54 @@
...
+		width: 200px;
+		height: 100px;
+		horizontal-align: center;
+		vertical-align: center;
+		overflow: noclip;
+	}
+
+	&__pitch {
+		horizontal-align: left;
+		margin-left: -40px;
```
tsa96, line 65: Negative margins are hacky and I try to only use them in cases with flow layout where they can be really useful. In this case `transform: translatex(-40px);` should work fine.

scripts/hud/cgaz.js
```diff
@@ -165,7 +162,19 @@ class Cgaz {
 		this.windicator_size = windicatorArrowConfig.size;
 		this.windicator_color = windicatorArrowConfig.color;
 
-		// TODO: apply arrow config
+		this.bShouldUpdateStyles = true;
+	}
+
+	static onCompassConfigChange() {
+		const CompassConfig = DefragAPI.GetHUDCompassCFG();
```
tsa96, line 169: Nit but should be lower CC since it's only in this method scope.

tsa96 (changes requested):
Do you have English POEditor terms ready? I can't remember if I've explained this already, but you can create a JSON file in this format (there's a template file in Localization folder on team storage)
```json
[
    {
        "term": "Settings_Defrag_Compass_Title",
        "definition": "Display compass"
    },
    {
        "term": "",
        "definition": ""
    },
    {
        "term": "",
        "definition": ""
    }
]
```
then in POEditor go to Import, select the JSON file, set to include definitions for English, and import. Can do that as soon as this is merged.

tsa96:
Also since I fixuped my misc-tweaks branch to fix the missing negation in gamemode check in cgaz.js, and it's fixed in your branch, you'll need to do a manual rebase

styles/hud/cgaz.scss
PeenScreeker: yea, good look. I was in margin mode from the other panels in this change. Agreed, updated to use translate

styles/hud/cgaz.scss
PeenScreeker: re: POEditor, I was poking around with the translations, but wasn't sure about the order of operations. I'm ready to upload the English definitions when this is merged.

tsa96: Yeah I think it's best to prep them at the same time as making the PR then import once PR is merged.

tsa96 (approved):

PR #45, opened 2022-12-06
Spelling Fixes

davis-matthew:
These are autogenerated comment spelling/grammar fixes. 
Two notes:
1. It was set to English-UK due to the spelling of initialise and such, so those inferior spellings are ignored.
2. This does not fix the fact that the methods like setComparisionStats() are misspelled, only the comments.


Gocnak (approved):

Gocnak:
Thank you for these!

PR #49, opened 2022-12-30
fix: Set ammo via dialog variable

AWildErin:
closes https://github.com/momentum-mod/game/issues/1955

Requires internal PR to be merged first

### DIFF
--- a/layout/hud/ammo.xml
+++ b/layout/hud/ammo.xml
@@ -4,6 +4,6 @@
 	</styles>
 
 	<MomHudAmmo class="full-width">
-		<Label class="ammo__label" id="AmmoCount" />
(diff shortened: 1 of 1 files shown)


tsa96 (changes requested):
This is a silly nit but we use lower_snake_case on dialog vars

AWildErin:
> This is a silly nit but we use lower_snake_case on dialog vars

Oop, didn't realise we did that! I've fixed that now!

tsa96:
This is camel but looking at stuff, we're not being consistent with the naming, so who cares lol

tsa96 (approved):

Gocnak (approved):

PR #54, opened 2023-01-20
fix: Chat spec icon

Mac-tf2:
Chat icon was using the wrong filename.


tsa96 (approved):
Legend, you are now a programmer!!

Gocnak (approved):

PR #57, opened 2023-01-22
Add sleep_when_meeting_framerate option to video settings

Arkanoid0:
```
"Settings_Video_PreventThreadSleep" "Prevent Thread Sleep"
"Settings_Video_PreventThreadSleep_Info" "Prevent the game thread from sleeping when at maximum FPS limit. Enabling this option results in higher CPU utilization and potentially smoother performance."
```


tsa96 (approved):
Nice work, looks good! One thing to keep in mind for any future PRs, try use the commit naming prefixes like `feat: ` etc., but I'll add a check for that in PR template soon.

PR #58, opened 2023-01-23
Linting and file structure refactor

tsa96:
Okay this one's a TON of commits and changes, though it's pretty much entirely refactors so shouldn't change behaviour.

Primarily adds a bunch of ESLint rules, using a pretty aggressive mixture of eslint/recommended and [unicorn](https://github.com/sindresorhus/eslint-plugin-unicorn) rules. I've kept them as similar as possible to the website repo.

Then fixes all those rules, some automatically, some manually.

Finally, there's the HUGE file structure refactor that also requires changes from my platform and red PRs - see red for details..


scripts/settings_speedometer.js
```diff
@@ -483,8 +483,11 @@ class RangeColorProfileObject {
 	}
 
 	createDisplayPanels(profileKV) {
-		Object.keys(this.displayObjectList)?.forEach((id) => this.displayObjectList[id]?.destroy());
+		const displayObjects = Object.keys(this.displayObjectList);
+		if (displayObjects) for (const id of displayObjects) this.displayObjectList[id]?.destroy();
```
Gocnak, line 487: Use null coalesce here like you do in a later commit.

scripts/hud/hud_leaderboards.js
```diff
@@ -52,10 +52,13 @@ class HudLeaderboards {
...
 	static setMapAuthorCredits(credits) {
 		// Delete existing name labels
-		this.panels.credits
-			.Children()
-			.slice(1) // Keep the "by" label
-			?.forEach((child) => child.DeleteAsync(0));
+		const existingLabels = this.panels.credits.Children().slice(1);
+
+		if (existingLabels) {
+			for (const label of existingLabels) {
```
Gocnak, line 58: Use null coalesce here as you do in a later commit.

Gocnak (changes requested):

scripts/settings_speedometer.js
tsa96: Done this and below in last push

Gocnak (approved):

PR #86, opened 2023-04-22
Feat: Improve optimal yaw delta calculation

5xp:
Closes momentum-mod/game/issues/2044

I've implemented the abstract optimal yaw calculations in Synchronizer, meaning it shows the optimal yaw with different acceleration/speed cap/friction values. Nothing changes with air strafing, but it does allow for Synchronizer to be useful when on the ground.

One note about this implementation is that `yawDelta` seems to break down at lower framerates.

### Checks


scripts/hud/synchronizer.js
```diff
@@ -291,6 +339,10 @@ class Synchronizer {
 		return Number.isNaN(Number(val)) ? def : val;
 	}
 
+	static FiniteCheck(val, def) {
```
tsa96, line 342: Method name should be lowerCamelCase

scripts/hud/synchronizer.js
```diff
@@ -31,32 +31,58 @@ class Synchronizer {
...
 			this.gainRatioHistory,
 			this.sampleWeight * this.NaNCheck(lastTickStats.speedGain / lastTickStats.idealGain, 0)
 		);
 		const gainRatio = this.getBufferedSum(this.gainRatioHistory);
 
-		const ratio = this.mom_hud_synchro_mode > 2 ? 1 - lastTickStats.yawRatio : lastTickStats.yawRatio;
-		this.addToBuffer(this.yawRatioHistory, this.sampleWeight * this.NaNCheck(ratio, 0));
+		const ratio =
+			this.mom_hud_synchro_mode > 2 ? 1 - Math.abs(yawDelta) / optimalDelta : Math.abs(yawDelta) / optimalDelta;
+		this.addToBuffer(this.yawRatioHistory, this.sampleWeight * this.FiniteCheck(ratio, 0));
```
tsa96, line 85: Since `finiteCheck` is so similar to `Number.isFinite` and only being used once, I don't think there's much point defining a separate function. So can just do `this.sampleWeight * (Number.isFinite(ratio) ? ratio : 0)`

scripts/hud/synchronizer.js
```diff
@@ -31,32 +31,58 @@ class Synchronizer {
...
 		const flip = this.mom_hud_synchro_flip_enable === 1 ? -1 : 1;
 
+		const yaw = MomentumPlayerAPI.GetAngles().y;
+		const yawDelta = bValidWishMove ? this.wrapValueToRange(this.lastYaw - yaw, -180, 180) * this.deg2rad : 0;
+		this.lastYaw = yaw;
+
+		lastMoveData.isGrounded = lastMoveData.acceleration !== this.airAcceleration; // lastMoveData.moveStatus doesn't work
+		lastMoveData.maxspeed = lastMoveData.isGrounded ? this.runSpeed : this.airSpeed;
+
+		const spd = lastMoveData.isGrounded ? this.applyFriction(this.maxGroundSpeed) : lastSpeed;
```
tsa96, line 74: What's `spd` stand for here, "speed" or something acronym ? Worth having a clearer name.

scripts/hud/synchronizer.js
```diff
@@ -31,32 +31,58 @@ class Synchronizer {
...
 		const strafeRight = (bValidWishMove ? 1 : 0) * lastTickStats.strafeRight;
 		const direction = this.mom_hud_synchro_dynamic_enable === 1 ? strafeRight : 1;
 		const flip = this.mom_hud_synchro_flip_enable === 1 ? -1 : 1;
 
+		const yaw = MomentumPlayerAPI.GetAngles().y;
+		const yawDelta = bValidWishMove ? this.wrapValueToRange(this.lastYaw - yaw, -180, 180) * this.deg2rad : 0;
+		this.lastYaw = yaw;
+
+		lastMoveData.isGrounded = lastMoveData.acceleration !== this.airAcceleration; // lastMoveData.moveStatus doesn't work
+		lastMoveData.maxspeed = lastMoveData.isGrounded ? this.runSpeed : this.airSpeed;
```
tsa96, line 72: Is `maxspeed` just some extra stuff you're adding to this object, or is it a value originally on `MomentumPlayerAPI`? If the latter, can just ignore this comment. But if it is a new value, I think it's clearer to keep this as a separate `const maxSpeed = lastMoveData.grounded ? ...` and add as an argument to `findOptimalDelta`. Since JS is untyped it's often difficult keeping track of what properties exist on what object, so best to avoid just chucking extra ones onto them.

scripts/hud/synchronizer.js
tsa96: Oh but I guess as per above comment, should be deleted anyway.

tsa96 (changes requested):
Mostly good, few small things. Reminder that when addressing changes please amend/fixup into your original commit then force-push, rather than separate commit.

Then best Peen reviews as well before we get this in since he knows the maths better than me.

scripts/hud/synchronizer.js
5xp: Yeah, `maxspeed` is a value originally on `MomentumMovementAPI.GetLastMoveData()` but it, along with `moveStatus` and `friction` apparently aren't hooked up and only work on Defrag.

5xp: I decided to make `onGround` its own variable and then pass that into `findOptimalDelta` instead of `lastMoveData` where we then determine `maxSpeed`. Though this shouldn't even matter because `lastMoveData` should have these values but they only work on Defrag. Should I make a separate issue for that?

PeenScreeker: Hey, I should have some time later today to review. The missing `lastMoveData` fields should be hooked up in engine. I can work on making those engine changes. Feel free to create an issue.

PeenScreeker:
These changes were implemented where stats are calculated in-engine.

Thank you for this pull request! I'm really looking forward to more creative UI applications from you. Feel free to reach out.

PR #90, opened 2023-06-13
Defrag Prime Sight

PeenScreeker:
Depends on Red changes for convars!

This pull request adds Prime Sight UI for Defrag. Prime Sight is a combination of the CGAZ and snap HUDs from Q3DF which shows the following information in real time:
- Which regions allow the player to gain speed
- How the potential gain from each zone compares with the maximum
- When it's favorable to switch strafe directions to maximize speed gain

Settings for the following features are included:
- Size (height) of the HUD tool
- Offset from center screen
- Minimum speed to draw the HUD
- Colors for indicating potential speed gain/loss
- Color for indicating diagonal strafe zones not currently in use
- Highlighting the zone representing the current movement calculation
- Color for highlighting current active zone
- Disabling the relative gain scaling for zone height
- Enabling an arrow to indicate the snapped direction of the current zone
- Size of the arrow
- Color of the arrow
** Note: trueness modifications (ground zones, A/D zones, jump/crouch) still WIP, but inaccessible from settings. The settings allow modification of fully complete features only.

### Checks

```json
[
  {
    "term": "Settings_Defrag_PrimeSight_Title",
    "definition": "Prime sight HUD"
  },
  {
    "term": "Settings_Defrag_PrimeSight_Enable",
    "definition": "Enable prime sight"
  },
  {
    "term": "Settings_Defrag_PrimeSight_Enable_Info",
    "definition": "Enable prime sight hud. This tool shows how movement inputs and view direction interact when strafing. To gain speed optimally, aim the crosshair in the tallest rectangle."
  },
  {
    "term": "Settings_Defrag_PrimeSight_Height",
    "definition": "Height"
  },
  {
    "term": "Settings_Defrag_PrimeSight_Height_Info",
    "definition": "Height to draw prime sight hud in pixels  @ 1080p"
  },
  {
    "term": "Settings_Defrag_PrimeSight_Offset",
    "definition": "Offset"
  },
  {
    "term": "Settings_Defrag_PrimeSight_Offset_Info",
    "definition": "Offset from center screen in pixels  @ 1080p"
  },
  {
    "term": "Settings_Defrag_PrimeSight_MinSpeed",
    "definition": "Minimum Speed"
  },
  {
    "term": "Settings_Defrag_PrimeSight_MinSpeed_Info",
    "definition": "Minimum speed threshold for drawing prime sight hud"
  },
  {
    "term": "Settings_Defrag_PrimeSight_GainColor",
    "definition": "Gain color"
  },
  {
    "term": "Settings_Defrag_PrimeSight_GainColor_Info",
    "definition": "Color to represent active zones that will cause the player's speed to increase on the next frame"
  },
  {
    "term": "Settings_Defrag_PrimeSight_LossColor",
    "definition": "Loss color"
  },
  {
    "term": "Settings_Defrag_PrimeSight_LossColor_Info",
    "definition": "Color to represent active zones that will cause the player's speed to decrease on the next frame"
  },
  {
    "term": "Settings_Defrag_PrimeSight_AltColor",
    "definition": "Inactive color"
  },
  {
    "term": "Settings_Defrag_PrimeSight_InactiveEnable",
    "definition": "Show inactive zones"
  },
  {
    "term": "Settings_Defrag_PrimeSight_InactiveEnable_Info",
    "definition": "Draw diagonal strafe zones for directions not currently active based on movement input"
  },
  {
    "term": "Settings_Defrag_PrimeSight_AltColor_Info",
    "definition": "Color to represent zones not currently active based on movement input"
  },
  {
    "term": "Settings_Defrag_PrimeSight_HighlightEnable",
    "definition": "Highlight current zone"
  },
  {
    "term": "Settings_Defrag_PrimeSight_HighlightEnable_Info",
    "definition": "Outline the current zone that will influence the player's movement calculation"
  },
  {
    "term": "Settings_Defrag_PrimeSight_HighlightColor",
    "definition": "Highlight color"
  },
  {
    "term": "Settings_Defrag_PrimeSight_HighlightColor_Info",
    "definition": "Color used to outline the current active zone"
  },
  {
    "term": "Settings_Defrag_PrimeSight_HighlightBorder",
    "definition": "Highlight border"
  },
  {
    "term": "Settings_Defrag_PrimeSight_HighlightBorder_Info",
    "definition": "Highlight border thickness in pixels"
  },
  {
    "term": "Settings_Defrag_PrimeSight_SizeByGain",
    "definition": "Scale height by gain"
  },
  {
    "term": "Settings_Defrag_PrimeSight_SizeByGain_Info",
    "definition": "Toggle scaling of zones by speed gain potential"
  },
  {
    "term": "Settings_Defrag_PrimeSight_ColorByGain",
    "definition": "Color zones by gain"
  },
  {
    "term": "Settings_Defrag_PrimeSight_ColorByGain_Info",
    "definition": "Toggle coloring of zones by speed gain potential"
  },
  {
    "term": "Settings_Defrag_PrimeSight_ArrowEnable",
    "definition": "Arrow"
  },
  {
    "term": "Settings_Defrag_PrimeSight_ArrowEnable_Info",
    "definition": "Draw an arrow to show the snapped direction of the current zone"
  },
  {
    "term": "Settings_Defrag_PrimeSight_ArrowSize",
    "definition": "Arrow size"
  },
  {
    "term": "Settings_Defrag_PrimeSight_ArrowSize_Info",
    "definition": "Size of the prime sight arrow. Value represents arrow half-height in pixels @ 1080p"
  },
  {
    "term": "Settings_Defrag_PrimeSight_ArrowColor",
    "definition": "Arrow color"
  },
  {
    "term": "Settings_Defrag_PrimeSight_ArrowColor_Info",
    "definition": "Color to draw the prime sight arrow"
  }
]
```

### DIFF
--- a/scripts/hud/cgaz.js
+++ b/scripts/hud/cgaz.js
@@ -148,9 +148,9 @@ class Cgaz {
 		this.applyClass(this.rightFastZone, FAST_CLASS);
 		this.applyClass(this.rightTurnZone, TURN_CLASS);
 		this.applyClass(this.accelSplitZone, NEUTRAL_CLASS);
-		this.applyClassBorder(this.leftMirrorZone, MIRROR_CLASS);
-		this.applyClassBorder(this.rightMirrorZone, MIRROR_CLASS);
-		this.applyClassBorder(this.mirrorSplitZone, MIRROR_CLASS);
+		this.applyClassBorder(this.leftMirrorZone, 2, MIRROR_CLASS);
+		this.applyClassBorder(this.rightMirrorZone, 2, MIRROR_CLASS);
+		this.applyClassBorder(this.mirrorSplitZone, 2, MIRROR_CLASS);
 
 		if (this.snapEnable) {
 			this.onSnapConfigChange();
@@ -213,7 +213,7 @@ class Cgaz {
 		);
 
 		WIN_ZONE_CLASS = new StyleObject(accelConfig.height, accelConfig.offset, this.windicatorColor);
-		this.applyClassBorder(this.windicatorZone, WIN_ZONE_CLASS);
+		this.applyClassBorder(this.windicatorZone, 2, WIN_ZONE_CLASS);
 	}
 
 	static onCompassConfigChange() {
@@ -804,10 +804,10 @@ class Cgaz {
 		panel.style.overflow = 'noclip noclip';
 	}
 
-	static applyClassBorder(panel, zoneClass) {
+	static applyClassBorder(panel, thickness, zoneClass) {
 		panel.style.height = this.NaNCheck(zoneClass.height, 0) + 'px';
-		panel.style.border = `2px solid ${zoneClass.color}`;
-		panel.style.padding = '-2px';
+		panel.style.border = `${thickness}px solid ${zoneClass.color}`;
+		panel.style.padding = `-${thickness}px`;
 		panel.style.verticalAlign = zoneClass.align;
 		panel.style.overflow = 'noclip noclip';
 	}
--- a/scripts/hud/cgaz.js
+++ b/scripts/hud/cgaz.js
@@ -718,9 +718,7 @@ class Cgaz {
 			this.updateZone(zones[i], left, right, 0, snapClass, this.snapSplitZone);
 
 			if (this.snapColorMode) {
-				const A = this.splitColorString(this.snapSlowColor);
-				const B = this.splitColorString(this.snapFastColor);
-				snapColor = this.getColorStringFromArray(this.colorLerp(A, B, alpha));
+				snapColor = this.colorLerp(this.snapSlowColor, this.snapFastColor, alpha);
 			}
 
 			let bHighlight = false;
@@ -940,8 +938,12 @@ class Cgaz {
 			.map((c, i) => (i === 3 ? Number.parseInt(c * 255) : Number.parseInt(c)));
 	}
 
-	static colorLerp(A, B, alpha) {
-		return A.map((Ai, i) => Ai + alpha * (B[i] - Ai));
+	static colorLerp(stringA, stringB, alpha) {
+		const arrayA = this.splitColorString(stringA);
+		const arrayB = this.splitColorString(stringB);
+		if (arrayA.length === 3) arrayA.push(255);
+		if (arrayB.length === 3) arrayB.push(255);
+		return this.getColorStringFromArray(arrayA.map((Ai, i) => Ai + alpha * (arrayB[i] - Ai)));
 	}
 
 	static {
--- a/scripts/hud/cgaz.js
+++ b/scripts/hud/cgaz.js
@@ -1,12 +1,27 @@
 'use strict';
 
 // TODO: remove these globals
-let MAX_GROUND_SPEED = 320;
+let MAX_GROUND_SPEED = 320; // initialized to 320. Changes with haste status.
 const AIR_ACCEL = 1;
 const DEFAULT_ACCEL = 2.56;
 const DEFAULT_SPEED = 320;
 const HASTE_ACCEL = 3.328; // (max speed) * (air accel = 1) * (tick interval) * (haste factor)
 const HASTE_SPEED = 416;
+
+const TruenessMode = {
+	GROUND: 1 << 0, // show ground zones
+	PROJECTED: 1 << 1, // show zones scaled by +jump/+crouch
+	CPM_TURN: 1 << 2 // show a/d and ground zones
+};
+
+const InputMode = {
+	NONE: 0,
+	AIR_CONTROL: 1,
+	STRAFE_LEFT: 2,
+	STRAFE_RIGHT: 3,
+	TURN_LEFT: 4,
+	TURN_RIGHT: 5
+};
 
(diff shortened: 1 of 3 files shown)


scripts/hud/cgaz.js
```diff
@@ -922,6 +1292,17 @@ class Cgaz {
 		return angle < 0 ? angle + Math.PI : angle - Math.PI;
 	}
 
+	static rotateVector(vector, angle) {
+		const out = JSON.parse(JSON.stringify(vector));
```
tsa96, line 1296: This is a hack that shouldn't be used lightly, in this case I'm sure it's faster to do
```js
const out = { x: vector.x, y: vector.y, z: vector.z };
```

but can't this function be simplified to just
```js
const cos = Math.cos(angle);
const sin = Math.sin(angle);

return { 
  x: vector.x * cos - vector.y * sin,
  y: vector.y * cos + vector.x * sin,
  z: vector.z // if there's a z component
}
```

scripts/hud/cgaz.js
```diff
@@ -944,6 +1325,11 @@ class Cgaz {
 		return A.map((Ai, i) => Ai + alpha * (B[i] - Ai));
 	}
 
+	static getRgbFromRgba(colorString) {
+		const colorArray = this.splitColorString(colorString);
+		return `rgb(${colorArray[0]}, ${colorArray[1]}, ${colorArray[2]})`;
```
tsa96, line 1330: Use destructuring:
```js
const [r, g, b] = this.splitColorString(colorString);
return `rgb(${r}, ${g}, ${b})`
```

tsa96 (changes requested):
Nice, looks like really clean code overall. Looking forward to trying this!

scripts/hud/cgaz.js
PeenScreeker: yea, tbh I absolutely hate the stringify/parse pattern to get a deep copy of an object. I did a little digging into the history of it and apparently V8 does this very well for small objects. I had entertained the idea of using `structuredClone` just because it's less gross but it's slower for small objects like this.

Your suggestion is the best way forward for this case. I'll make the change to remove the `out` variable.

tsa96: Haha yeah Javascript deep copying is a real rabbit hole. You're right, V8 has some optimisations specifically because the stringify-parse method is so common, I still find it very ugly to see it written though, and crucially only works for a subset of Javascript objects, no functions. Honestly with browsers/node I'd probably always use structuredClone for nested objects as perf may improve in future.

On that - deep copying isn't even needed here. It's only needed for nested objects, these values only have primitive values. So you could've just done a regular shallow copy with `out = { ...vector };`.

Lastly, worth noting that we don't actually have structuredClone. It's part of some web API spec so browsers implement it individually, and Node has it. I've looked into doing it ourselves, even just doing the parse-stringify stuff from C++ which is a bitttttt cleaner, but doing a proper structuredClone implementation would probably be best. Just requires writing an object traversal algorithm (can't remember what structuredClone uses, but I think it's in the spec) and handling all the cursed V8 C++ API object persistence stuff.

tsa96 (approved):

chovelyoukai:
Trying this out locally, seems like the `mom_df_hud_prime_hl_enable` doesn't have an effect, it's always enabled regardless of the cvar setting. Looks like this isn't hooked up anywhere?

scripts/hud/cgaz.js
```diff
@@ -192,6 +220,58 @@ class Cgaz {
...
+		const primeConfig = DefragAPI.GetHUDPrimeCFG();
+		this.primeEnable = primeConfig.enable;
+		this.primeTruenessMode = primeConfig.truenessMode;
+		this.primeMinSpeed = primeConfig.minSpeed;
+		this.primeHeight = primeConfig.height;
+		this.primeOffset = primeConfig.offset;
+		this.primeGainColor = primeConfig.gainColor;
+		this.primeLossColor = primeConfig.lossColor;
+		this.primeAltColor = primeConfig.altColor;
+		this.primeHlenable = primeConfig.highlightEnable;
```
chovelyoukai, line 233: nit: "primeHlEnable"

chovelyoukai:
Appears to not work with half beat and dual beat in VQ3, I don't see accel zones when holding diagonal keys.

In vintage, the accel zones show up when holding diagonal keys as if it's Q3 strafing, even though in vintage it's still QW strafing.

chovelyoukai:
Copying suggestions from team meeting:
1. Add a cvar to turn off "mirror" zones so the player can just press a button in settings to turn them off, instead of making them fully transparent.
2. Consider turning off "mirror" zones by default to match cgaz defaults, but get some more opinions from df players. If the majority of people have them on and use them, then they should be on by default.
3. Styling wise, I think it looks good. I recommend matching cgaz defaults wherever it makes sense.

scripts/hud/cgaz.js
PeenScreeker: nice catch

PeenScreeker:
> 

Added setting to show/hide inactive diagonal strafe zones (cgaz equivalent to mirror zones), default off.

chovelyoukai (approved):

PR #91, opened 2023-06-21
feat: add cm/360 sensitivity conversion

Natanxp2:
Closes momentum-mod/game/issues/2047

This PR adds conversion from in-game sensitivity to cm/360 and in/360 ( and vice-versa )

### Checks

```json
[
  {
    "term": "Settings_DPI_Input",
    "definition": "Input DPI"
  },
  {
    "term": "Settings_Centimeters_Per_360",
    "definition": "cm/360"
  },
  {
    "term": "Settings_Inches_Per_360",
    "definition": "in/360" 
  }
]
```

### DIFF
--- a/scripts/pages/settings/mouse-sensitivity.js
+++ b/scripts/pages/settings/mouse-sensitivity.js
@@ -1,0 +1,58 @@
+'use strict';
+
+const CM_CONVERSION_FACTOR = 2.54;
+
+class MouseSensitivity {
+	static panels = {
+		/** @type {TextEntry} @static */
+		dpi: $('#DPI'),
+		/** @type {DropDown} @static */
+		unitSelection: $('#UnitSelection'),
+		/** @type {TextEntry} @static */
+		unitsPer360: $('#UnitsPer360')
+	};
+
+	static calculateSensitivity() {
+		if (!this.panels.dpi.text) return;
+
+		const yaw = GameInterfaceAPI.GetSettingFloat('m_yaw');
+		const eDPI = this.panels.dpi.text * GameInterfaceAPI.GetSettingFloat('sensitivity');
+		let per360 = 360 / (eDPI * yaw);
+
+		if (this.panels.unitSelection.GetSelected().id === 'm_sens_cm') per360 *= CM_CONVERSION_FACTOR;
+
+		this.panels.unitsPer360.text = per360.toFixed(2);
+	}
+
+	static setSensitivity() {
+		if (!this.panels.dpi.text) return;
+
+		const yaw = GameInterfaceAPI.GetSettingFloat('m_yaw');
+		const eDPI = 360 / yaw / this.panels.unitsPer360.text;
+		let sensitivity = eDPI / this.panels.dpi.text;
+
+		if (this.panels.unitSelection.GetSelected().id === 'm_sens_cm') sensitivity *= CM_CONVERSION_FACTOR;
+
+		GameInterfaceAPI.SetSettingFloat('sensitivity', sensitivity);
+	}
+
+	static saveDPI() {
+		$.persistentStorage.setItem('settings.mouseDPI', this.panels.dpi.text);
+		this.calculateSensitivity();
+	}
+
+	static saveUnitSelection() {
+		$.persistentStorage.setItem('settings.mouseUnitSelector', this.panels.unitSelection.GetSelected().id);
+		this.calculateSensitivity();
+	}
+
+	static loadSettings() {
+		this.panels.dpi.text = $.persistentStorage.getItem('settings.mouseDPI');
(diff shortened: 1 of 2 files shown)


scripts/pages/settings/mouse-sensitivity.js
```diff
@@ -0,0 +1,63 @@
...
+	
+	static saveUnitSelection(){
+		$.persistentStorage.setItem('settings.unitSelection', this.panels.unitSelection.GetSelected().id);
+		this.calculateSensitivity();
+	}
+	
+	static loadSettings(){
+		this.panels.dpi.text = $.persistentStorage.getItem('settings.dpi');
+		this.panels.unitSelection.SetSelected($.persistentStorage.getItem('settings.unitSelection'));
+		this.calculateSensitivity();
```
tsa96, line 56: These PS keys are a little unclear, I suggest `settings.mouseDPI` and `settings.mouseUnitSelector`

scripts/pages/settings/mouse-sensitivity.js
```diff
@@ -0,0 +1,63 @@
...
+		const yaw = GameInterfaceAPI.GetSettingFloat('m_yaw')
+		const eDPI = this.panels.dpi.text * GameInterfaceAPI.GetSettingFloat('sensitivity');
+		let per360 = (360 / (eDPI * yaw));
+
+		if(this.panels.unitSelection.GetSelected().id === 'm_sens_cm')
+		per360 *= CM_CONVERSION_FACTOR;
+		
+		//displayprecision attribute does not work on TextEntry
+		per360 = Math.round(per360 * 100)/100;
+		this.panels.unitsPer360.text = per360;
```
tsa96, line 27: Bit cleaner to use `toFixed` here, so `this.panels.unitsPer360.text = per360.toFixed(2);`.

styles/pages/settings/settings.scss
```diff
@@ -103,6 +103,15 @@ $page-width: 1920px - $rightnav-width;
...
 		border-radius: 2px;
 	}
+	
+	&__sensitivity {
+		padding: 0 6px 0 12px;
+		margin-bottom: 4px;
+		
+		&--margin-left {
+			margin-left: 12px;
+		}
```
tsa96, line 113: I think it's best to get rid of this class, and just use a utility class (`ml-4`) is equivalent. Especially since this isn't actually proper BEM, since `--foo` means a modifier on an existing block/element, but you're using this classes on the outer `<Panel class="settings__sensitivity settings-enum__values">` and it's children. If you wanted to do properly in BEM you'd need to define a new block e.g. `settings-sensitivity` and do some element `settings-sensitivity__setting` or something. But just using `mb-4` is way simpler here.

tsa96 (changes requested):
Great work! This is a very reasonable implementation, glad you wrote a new JS class for this, better than bunging into existing files like we and Valve have done in the past.

Few comments, also you've not run Prettier so code formatting checks have failed. You need to install NodeJS and run `npm install` in the topmost dir, then `npm run format:fix` will format everything. Make sure you just amend that and addressing below comments into your existing commit (see contribution guidelines for this repo and the gh org for more info).

For the sensitivity change event, we've been discussing internally today and the plan is to add a more generic `$.RegisterConvarListener` or similar that can register a callback with any convar change. Not totally sure when that'll get in (I don't have time to work on it atm, sorry!), we'll discuss it more on Saturday.

Natanxp2:
I updated the code to address the comments. Regarding using utility classes, code is cleaner now but there is no 12px spacer defined. I have used 16px instead. I don't think this matters too much but if 12px is preferred let me know, I'll do it with a proper BEM then.

tsa96:
Yeah that's good with me. SCell has written a `$.RegisterConvarListener` handler system already so hopefully we'll get that in sometime soon!

tsa96 (approved):

PeenScreeker:
got the convar listener working, this is what the pano syntax will look like:
`$.RegisterConVarChangeListener('sensitivity', this.calculateSensitivity.bind(this));`

Gocnak:
The convar listener change is being merged in for 0.9.15, you can resume work after that release 👍

tsa96:
Big props to Peen and SCell for getting this into engine so fast!

Natanxp2:
I added the convarListener. I know that it passes the cvar to a bound function but since I have to calculate sensitivity also when DPI is changed it seems pointless to do something like 
`const sensitivity = <passed variable> || GameInterfaceAPI.GetSettingFloat('sensitivity');`
Using only the API seems cleaner to me.

tsa96:
> I added the convarListener. I know that it passes the cvar to a bound function but since I have to calculate sensitivity also when DPI is changed it seems pointless to do something like `const sensitivity = <passed variable> || GameInterfaceAPI.GetSettingFloat('sensitivity');` Using only the API seems cleaner to me.

Yes, agreed

PR #107, opened 2023-11-28
Move util functions out of cgaz.js

PeenScreeker:
This pull request moves math functions out of the cgaz.js file to separate util files `math.js` and `colors.js`. The intent is to move these out for:
- reducing mental load reading cgaz.js
- prep to make cgaz modular components (like all other hud tools)
- potential reuse by other contributors (color functions, and mapping angles to the screen)

Upon merging this PR, cgaz.js should be broken up into separate tools. It seemed appropriate to break this into two separate pull requests to keep the refactor organized.

### Checks


scripts/util/colors.js
```diff
@@ -0,0 +1,71 @@
...
+ * @param {number} alpha
+ * @returns {string}
+ */
+function colorLerp(colorA, colorB, alpha) {
+	const arrayA = splitColorString(colorA);
+	const arrayB = splitColorString(colorB);
+	const interp = Math.max(Math.min(alpha, 1), 0);
+	if (arrayA.length === 3) arrayA.push(255);
+	if (arrayB.length === 3) arrayB.push(255);
+	return getColorStringFromArray(arrayA.map((Ai, i) => Ai + interp * (arrayB[i] - Ai)));
```
tsa96, line 47: Okay this confused me and made me yearn for Typescript. Check out my post in #panorama, I've rewritten this file for you.

scripts/util/math.js
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * 2D length of input vector
+ * @param {object} vec
+ * @returns {number}
+ */
+function getSize(vec) {
+	return Math.sqrt(getSizeSquared(vec));
+}
+
```
tsa96, line 14: This is worth doing in TS as well, I think it's so self-explanatory you should be able to do it even if you've not used TS before. You just need
```ts
type Vec2D = [number, number];
type Vec3D = [number, number, number];
```
though maybe worth porting those in `index.d.ts`, that'd need a separate PR into the submodule though so I guess just leave here for now. You can remove JSDoc type annoations if the args and return types are typed.

scripts/util/math.js
```diff
@@ -0,0 +1,141 @@
+/**
+ * Utility functions for Javascript.
+ * Could be ported to C++ and exposed globally in the future.
+ */
+
```
tsa96, line 5: Make specific to this file

tsa96 (changes requested):
Since these are global utilities now, we should absolutely get them over to TypeScript. I've done `colors.ts` for you, `math.ts` should be very easy.

scripts/util/math.js
PeenScreeker: Converted these. Very straightforward

scripts/util/colors.js
PeenScreeker: Thanks for sharing the rewrite! I made a couple changes to handle converting `rgb` and `rgba` strings to tuples.

scripts/util/colors.js
```diff
@@ -0,0 +1,71 @@
...
+ * rgba(R, G, B, A)
+ * where R, G, B values ranged [0, 255], A ranged [0, 1].
+ * @param {array} color
+ * @returns {string}
+ */
+function splitColorString(string) {
+	return string
+		.slice(5, -1)
+		.split(',')
+		.map((c, i) => (i === 3 ? Number.parseInt(c * 255) : Number.parseInt(c)));
```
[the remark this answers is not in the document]
tsa96: Gotcha, good point. The thing I was pointing out (sorry if wasn't clear) is that `c` is a string until you parse it, so the `*` will concatenate, not multiply.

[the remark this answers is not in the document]
PeenScreeker: Yea, the comment was actually helpful! When I was debugging the lerp turning panels invisible, I ran into type errors trying to do the multiply inside. Probably put me on the right track faster to see the comment here than stumbling through on my own, tbh

scripts/util/colors.ts
```diff
@@ -0,0 +1,150 @@
...
+ * Blends two colors linearly (not HSV lerp).
+ * RGB inputs are converted to RGBA with A value of 1.
+ */
+function rgbaStringLerp(colorA: string, colorB: string, alpha: number, useHsv: boolean = false): string {
+	const arrayA = rgbaStringToTuple(colorA);
+	const arrayB = rgbaStringToTuple(colorB);
+	if (!useHsv) return tupleToRgbaString(rgbaTupleLerp(arrayA, arrayB, alpha));
+
+	const FromHSV: RgbaTuple = LinearRGBToHSV(arrayA) as RgbaTuple;
+	const ToHSV: RgbaTuple = LinearRGBToHSV(arrayB) as RgbaTuple;
```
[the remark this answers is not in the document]
tsa96: Bunch of off names in this file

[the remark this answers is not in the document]
PeenScreeker: lol, wasn't quite done. Should be good to go now

scripts/util/colors.ts
```diff
@@ -0,0 +1,153 @@
+/**
+ * Functions for manipulating RGB/RGBA strings and tuples.
+ */
+
+/** */
+type RgbaTuple = [number, number, number, number];
```
tsa96, line 6: Leave comment saying ranges

scripts/util/math.ts
```diff
@@ -0,0 +1,119 @@
...
+function getSizeSquared2D(vec: Vector): number {
+	return vec.x * vec.x + vec.y * vec.y;
+}
+
+/**
+ * Returns copy of input vector scaled to length 1.
+ * If input vector length is less than threshold,
+ * the zero vector is returned.
+ * @param {object} vec
+ * @returns {object}
```
tsa96, line 30: Remove

tsa96 (changes requested):

scripts/util/colors.ts
tsa96: [3] is ranged [0, 1], no?

tsa96 (changes requested):

scripts/util/colors.ts
PeenScreeker: In tuple form, all values are `[0, 255]`. I think it's that way because that's how the colors are set (as cvars, and in color picker). Looking at the code, it could be reworked for `[0, 1]`, but the current way allows me to treat all the numbers the same for processing instead of dealing with both ints and a float.

PeenScreeker: hey @tsa96 I just realized this comment reply didn't populate on the main conversation tab. Just pinging you incase it slipped through the cracks

tsa96: Woops, sorry!

tsa96 (approved):

tsa96:
Sorry didn't realise I hadn't resolved everything

PR #125, opened 2024-04-06
refactor: Misc fixes for API model updates

Panzerhandschuh:
- Use `UserMapData` for checking if the map file exists or is in favorites
- Lookup main track leaderboard since it no longer exists on the model
- Updated map data to be compatible with new backend models

### DIFF
--- a/scripts/common/map-cache.js
+++ b/scripts/common/map-cache.js
@@ -20,13 +20,13 @@ const MapCreditType = {
 	SPECIAL_THANKS: 3
 };
 
-function GetMainTrack(mapData, gamemode) {
+function getMainTrack(mapData, gamemode) {
 	return mapData.leaderboards.find(
 		(leaderboard) =>
 			leaderboard.gamemode === gamemode && leaderboard.trackType === TrackType.MAIN && leaderboard.style === 0
 	);
 }
 
-function GetNumZones(mapData) {
+function getNumZones(mapData) {
 	return mapData.leaderboards.filter((leaderboard) => leaderboard.trackType === TrackType.STAGE).length;
 }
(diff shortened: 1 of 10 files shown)


tsa96:
JS looks fine besides not having autoformatting applied. CI should be yelling at you but the job is only configured to run on one branch.

My last PR enables CI everywhere, then https://github.com/momentum-mod/panorama/pull/127 brings the 0.10 branch up to date with that, please could you approve and merge that, then rebase this branch onto this. Then run `npm run format:fix` and amend those changes into this commit, then force-push. 

I'll double check this when more awake tomorrow and have reviewed the red side of this stuff

scripts/common/map-cache.js
```diff
@@ -0,0 +1,32 @@
...
+function GetMainTrack(mapData, gamemode) {
+	return mapData.leaderboards.find(
+		(leaderboard) =>
+			leaderboard.gamemode === gamemode && leaderboard.trackType === TrackType.MAIN && leaderboard.style === 0
+	);
+}
+
+function GetNumZones(mapData) {
+	return mapData.leaderboards.filter((leaderboard) => leaderboard.trackType === TrackType.STAGE).length;
+}
```
tsa96, line 32: Naming convention for functions is lowerCamelCase!

scripts/pages/map-selector/map-entry.js
```diff
@@ -28,84 +28,73 @@ class MapEntry {
 		const mapData = $.GetContextPanel().mapData;
```
tsa96, line 28: `$.GetContextPanel()` has to call into C++ every time which is nice to avoid. In this case I would just do 
```js
const { mapData, userMapData, isDownloading } = $.GetContextPanel()
```

tsa96 (changes requested):

PR #131, opened 2024-07-02
prepare tab menu for upcoming in-game zoning UI

PeenScreeker:
This pull request adds the requisite panels as skeletons for the upcoming zoning UI. Breaking changes were introduced in the engine-side code necessary for zone editing and UI. The empty panorama code introduced in this PR will be filled with zoning UI implementation in a future PR. The purpose of this MR is to add frontend changes necessary to prevent the 0.10 branch from breaking when merging such that the zoning UI can be tested on the 0.10 branch - potentially without red access.

Must be merged simultaneously with Red MR.

### DIFF
--- a/layout/hud/tab-menu.xml
+++ b/layout/hud/tab-menu.xml
@@ -51,6 +51,10 @@
 				<Frame src="file://{resources}/layout/pages/end-of-run/end-of-run.xml" hittest="true" hittestchildren="true" />
 			</Panel>
 
+			<Panel id="ZoningContainer" class="hud-tab-menu__zoning" hittestchildren="true">
+				<ZoneMenu id="ZoneMenu" />
+			</Panel>
+
 			<Panel class="hud-tab-menu__stats">
 			</Panel>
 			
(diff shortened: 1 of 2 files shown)


layout/hud/tab-menu.xml
```diff
@@ -51,6 +51,10 @@
 				<Frame src="file://{resources}/layout/pages/end-of-run/end-of-run.xml" hittest="true" hittestchildren="true" />
 			</Panel>
 
+			<Panel id="ZoningContainer" class="hud-tab-menu__zoning" hittestchildren="true">
```
tsa96, line 54: Isn't `hittestchildren` redundant here, since it's immediate child has `hittest` and `hittestchildren`? Don't know exactly how this works in code.

tsa96 (changes requested):

layout/hud/tab-menu.xml
PeenScreeker: could be changed, the user can't interact with this in any way currently. I copied this line from my zoning UI branch.

tsa96 (approved):

Panzerhandschuh (approved):

PR #144, opened 2024-08-26
Implemented status information for new timer

Panzerhandschuh:
- Display current run time
- Timer states (disabled, prime, running, finished)
- Track progress (current bonus, stage, checkpoint)
- Timer sounds (start, stop, finish)

Comparisons still need to be implemented.


PeenScreeker (approved):
lgtm

scripts/hud/timer.js
```diff
@@ -141,6 +138,65 @@ class HudTimer {
...
+				this.timeLabel.AddClass(INACTIVE_CLASS);
+				this.timeLabel.RemoveClass(FINISHED_CLASS);
+				break;
+			default:
+				$.Warning('Unknown timer state');
+				break;
+		}
+
+		$.GetContextPanel().SetDialogVariableFloat('runtime', MomentumTimerAPI.GetObservedTimerStatus().runTime);
+	}
```
tsa96, line 176: There's no reason to call into C++ every frame like this, just store the value of `$.GetContextPanel()` on the class.

tsa96 (changes requested):
Shouldn't we be merging this into the 0.10.0 branch?

tsa96 (approved):

PR #156, opened 2024-11-01
New run comparison API and set from leaderboard entry

jason-e:


tsa96 (changes requested):
LGTM, just need to run Prettier (`npm run format:fix`)

Panzerhandschuh (approved):

tsa96:
@jason-e pls let CI pass before merging! Just means I have to make extra commits fixing ESLint errors. Most editors have ESLint extensions/plugins that will autodetect our config, just need to install one.

PR #157, opened 2024-11-15
0.10 Map Selector Part 1

tsa96:
All the Panorama changes corresponding to #2345 on red. Some UI tweaks as well though still slightly WIP, want to get a PR in at a stable point corresponding to the incoming changes on red.

I'll have a subsequent branch on red and here open ASAP, C++ side of that is mostly finished.

### Checks

Made some POEditor additions recently but just added them directly.


tsa96:
Dunno what's up with ESLint will fix tomorrow

scripts/pages/map-selector/map-selector.ts
```diff
@@ -2,6 +2,8 @@ import { OnPanelLoad, PanelHandler } from 'util/module-helpers';
 import { traverseChildren } from 'util/functions';
 import { MapCreditType } from 'common/web';
 
+const DASHBOARD_URL = 'https://dashboard.momentum-mod.org';
+
```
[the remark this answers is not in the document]
tsa96: Yeah, I noticed this, and was doing in the second PR. Since afaik this hasn't been reviewed much and I have a second PR worth of stuff that's very related, I'm going to merge them together and fixup some commits so this is clearer.

scripts/common/leaderboard.ts
```diff
@@ -32,13 +32,36 @@ export enum LeaderboardStatusType {
...
+			leaderboard.style === style
 	);
 }
 
 export function getNumZones(mapData: MMap): number {
 	return mapData.leaderboards.filter((leaderboard) => leaderboard.trackType === TrackType.STAGE).length;
 }
+
+export function getNumBonuses(mapData: MMap): number {
+	return mapData.leaderboards.filter((leaderboard) => leaderboard.trackType === TrackType.STAGE).length;
```
Panzerhandschuh, line 56: This should be `TrackType.BONUS`

tsa96: Yeah this is fixed already, just haven't pushed yet. I'd hold off on reviewing today, I don't have the energy to get this branch finalized tonight and there's a lot missing.

Also note that this function is wrong in that it can count multiple gamemodes on the same track, same issue with C++.

tsa96: Fixed now

tsa96:
Reworked a bunch of stuff, left out gallery changes, though the main map image background is now set in TS rather than C++, since that works a lot better if we go down the in-selector map image route - whatever approach we go with, gallery stuff should be doable with only TS changes.

tsa96:
Sorry, I pushed these changes to a second branch a few days ago that isn't PRed, forgetting that I was going to just combine the two branches together. I know the PR gets bigger but there's a lot fewer changes than if they were split up since I fixuped so much.

Gocnak (approved):

scripts/pages/map-selector/map-selector.ts
```diff
@@ -504,4 +505,100 @@ class MapSelectorHandler implements OnPanelLoad {
 		this.panels.leaderboardContainer.SetHasClass('mapselector-leaderboards--open', open);
 		$.persistentStorage.setItem('mapSelector.leaderboardsOpen', open);
 	}
+
+	checkingUpdates = false;
+	lastUpdateCheck = 0;
+
+	checkForUpdates() {
+		if (this.checkingUpdates || this.lastUpdateCheck + REFRESH_COOLDOWN > Date.now()) {
```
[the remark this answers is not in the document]
tsa96: I was under the impression our plan was to configure rate-limiting in NGINX, which has less overhead than doing it per-endpoint in NestJS. Whilst this *is* quite an expensive query, it's probably not the most expensive, and someone trying to spam us deliberately would do a much better job picking a nastier query and calling it from a HTTP client than via Panorama.

I'd split the concerns here into two camps: innocent, impatient users spamming a button for the hell of it, and malicious users spamming it to try to overload our servers. This check is only relevant to the former camp, and makes no attempt to stop malicious users. Stopping those in the latter camp is very important, but I don't think we should put special measures in place specifically for people maliciously using this UI - there's better ways to spam us than this, and we should have better guards against it than rate limiting individual endpoints.

PR #160, opened 2024-11-26
Timer splits for stages and ordered/required checkpoints

Panzerhandschuh:
I will come back to this and add support for unordered/optional checkpoints and run stats comparisons.

### Checks

### DIFF
--- a/scripts/common/timer.ts
+++ b/scripts/common/timer.ts
@@ -362,10 +362,11 @@ export class Comparison {
 			accumulateTime: baseAccumulateTime,
 			time: baseSplitTime,
 			diff: baseAccumulateTime - comparisonAccumulateTime,
-			delta: baseSplitTime - comparisonSplitTime
-			// statsComparisons: baseRunSplits.trackStats.map(
-			// 	(stat, j) => new RunStatsComparison(stat, compareZone.stats[j])
-			// )
+			delta: baseSplitTime - comparisonSplitTime,
+			statsComparisons: generateStatsComparison(
+				baseRunSplits.segments[segmentIndex].subsegments[subsegmentIndex].stats,
+				comparisonRunSplits.segments[segmentIndex].subsegments[subsegmentIndex].stats
+			)
 		};
 	}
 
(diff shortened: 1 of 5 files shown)


PeenScreeker (approved):

PeenScreeker:
lgtm

tsa96 (changes requested):

tsa96:
This might not fit into current code perfect but for `RunStatsComparison` you could do something like this: 
```ts
const RunStatsUnits: Record<string, string> = {
	maxOverallSpeed: 'ups',
	maxHorizontalSpeed: 'ups',

	overallDistanceTravelled: 'units',
	horizontalDistanceTravelled: 'units',

	jumps: 'jumps',
	strafes: 'strafes'
};

export interface RunStatsComparison {
	name: string;
	unit: string;
	baseValue: number;
	comparisonValue: number;
	diff: number;
}

export function GenerateStatsComparison(baseStats: RunStats, comparisonStat: RunStats): RunStatsComparison[] {
	return (Object.keys(baseStats) as Array<keyof RunStats>).map((key) => ({
		name: key,
		unit: RunStatsUnits[key],
		baseValue: baseStats[key],
		comparisonValue: comparisonStat[key],
		diff: baseStats[key] - comparisonStat[key]
	}));
}
```

tsa96 (approved):
Was trying to understand what was going on in the constructor of the Comparison class and realised it's never even getting used.

We shouldn't need to keep Run_OLD around given the new RunSplits stuff, and so I don't really see the point of this class. The only thing that's still obviously relevant are generateSplits (though never gets used atm) and generateSubsegmentSplit. These are static methods and don't really have anything to do with comparisons, so they should just be plain functions.

We could pull this all out of classes and use constructors, but that's absolutely not a requirement in JS/TS, and since we have several instances of constructor-like functions that return arrays of what they should be constructing, it seems easier to skip using classes and just use interfaces + pure functions. We can then name the import of common/timer.ts in some places to scope the functions better.

Sorry to piggy-back your work but I *really* want to get the data structures organised well here and avoid using any of these unnecessary old classes. I've just spent a half hour reorganising stuff a little doing the WIP refactoring of the endofrun code. Let's merge what you have here, then I'm gonna push my WIP stuff on top of that, and I'd appreciate it if you or anyone else worked off of there when doing end of run

PR #166, opened 2024-12-23
Leaderboard fixes

Panzerhandschuh:
### Checks

### DIFF
--- a/scripts/pages/leaderboards/leaderboards.ts
+++ b/scripts/pages/leaderboards/leaderboards.ts
@@ -1,7 +1,7 @@
 import { exposeToPanelContext, PanelHandler } from 'util/module-helpers';
 import { LeaderboardListType, LeaderboardStatusType, LeaderboardType, sortLeaderboard } from 'common/leaderboard';
 import { EndOfRunShowReason } from 'common/timer';
-import { Leaderboard, MMap, TrackType } from 'common/web';
+import { MMap, TrackType } from 'common/web';
 
 exposeToPanelContext({ LeaderboardListType, LeaderboardType });
 
--- a/scripts/pages/leaderboards/leaderboards.ts
+++ b/scripts/pages/leaderboards/leaderboards.ts
@@ -41,9 +41,13 @@ class LeaderboardsHandler {
 		$.RegisterEventHandler('Leaderboards_TimesFiltered', $.GetContextPanel(), (count) =>
 			this.onTimesUpdated(count)
 		);
-		$.RegisterForUnhandledEvent('EndOfRun_Show', (reason) => this.onShowEndOfRun(reason));
-		$.RegisterForUnhandledEvent('Leaderboards_MapDataSet', (isOfficial) => this.onMapLoad(isOfficial));
-		$.RegisterForUnhandledEvent('Leaderboards_MapLeaderboardsLoaded', (map) => this.onMapLeaderboardsLoad(map));
+		$.RegisterEventHandler('EndOfRun_Show', $.GetContextPanel(), (reason) => this.onShowEndOfRun(reason));
+		$.RegisterEventHandler('Leaderboards_MapDataSet', $.GetContextPanel(), (isOfficial) =>
+			this.onMapLoad(isOfficial)
+		);
+		$.RegisterEventHandler('Leaderboards_MapLeaderboardsLoaded', $.GetContextPanel(), (map) =>
+			this.onMapLeaderboardsLoad(map)
+		);
 
 		// Note: Can't set radio button groups in the XML because it causes multiple leaderboard instances to interfere with eachother
 		const lbType = this.panels.cp.id === 'TabMenuLeaderboards' ? 'TabMenu' : 'MapSelector';
--- a/scripts/pages/leaderboards/leaderboards.ts
+++ b/scripts/pages/leaderboards/leaderboards.ts
@@ -194,7 +194,7 @@ class LeaderboardsHandler {
 			// Main track
 			{
 				const trackStr = $.Localize('#Leaderboards_Tracks_Main');
-				const item = $.CreatePanel('Label', this.panels.tracksDropdown, trackStr, {
+				const item = $.CreatePanel('Label', this.panels.tracksDropdown, '', {
 					text: trackStr,
 					value: 0
 				});
@@ -209,7 +209,7 @@ class LeaderboardsHandler {
 			if (segments.length > 1) {
 				segments.forEach((_, index) => {
 					const trackStr = `${$.Localize('#Leaderboards_Tracks_Stage')} ${index + 1}`;
-					const item = $.CreatePanel('Label', this.panels.tracksDropdown, trackStr, {
+					const item = $.CreatePanel('Label', this.panels.tracksDropdown, '', {
 						text: trackStr,
 						value: index + 1
 					});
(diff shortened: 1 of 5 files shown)


scripts/pages/leaderboards/leaderboards.ts
```diff
@@ -25,21 +25,42 @@ class LeaderboardsHandler {
...
 
 		// Default to Top 10
 		this.setSelectedTimesList(LeaderboardListType.GLOBAL);
 		this.setSelectedListType(LeaderboardListType.GLOBAL, LeaderboardType.TOP10);
 		this.setSelectedListType(LeaderboardListType.LOCAL, LeaderboardType.LOCAL);
+
+		this.panels.tracksDropdown.RemoveAllOptions();
+		this.panels.tracksDropdown.visible = false;
+
+		// Fix map selector radio button group state conflicting with tab menu
```
Panzerhandschuh, line 44: @tsa96 do you know if there's a way to assign dynamic values to the radio button groups in the XML instead of in TS? This code executes too late and the tab menu still changes the map selector leaderboard radio button state.

tsa96: AFAIK this is just a limitation of `RadioButton`s, it's gross how `group`s are completely global but not worth changing; fine to do in TypeScript. But let's stick with our usable approach of storing references to panels we access with `$` in the `panels` field, then you can iterate through them easily. So

```ts
readonly panles = {
  ...,
  radioButtons: {
    listTypes: {
      global: $.<RadioButton>('#TimesListGlobal')
      ...
    },
    local: {
      ... you get the idea
    },
    online: {
      ...
    }
  }
}
```
then in ctor
```ts
this.panels.listTypes.global.SetSelected(true);
this.panels.global.top10.SetSelected(true);
this.panels.local.runs.SetSelected(true);
```
then
```ts
// Note: Can't set radio button groups in the XML because it causes multiple leaderboard instances to interfere with eachother
const lbType = this.panels.cp.id === 'TabMenuLeaderboards' ? 'TabMenu' : 'MapSelector';
Object.entries(this.panels.radioButtons).forEach(([group, buttons] =>
  Object.values(buttons).forEach(button => (button.group = lbType + group))
)
```
groups aren't need in code anywhere else, just used by Pano internally, so group names aren't a big deal, using camelcase `listTypes` is fine.

scripts/pages/leaderboards/leaderboards.ts
```diff
@@ -272,6 +273,14 @@ class LeaderboardsHandler {
 		this.initTracksDropdown();
 	}
 
+	leaderboardTrackIdSort(a: Leaderboard, b: Leaderboard): number {
+		if (a.trackType === b.trackType) {
+			return a.trackNum - b.trackNum;
+		}
+
+		return a.trackType - b.trackType;
+	}
```
tsa96, line 282: If you're only using this in one place I think it's *much* cleaner to inline a sort fn like this. But likely we're use this again in the future anyway - iirc I wrote something like this recently, though don't remember if in website or panorama code. I'd pull it out to `common/leaderboard` and name `sortTrackId`.

scripts/pages/leaderboards/leaderboards.ts
```diff
@@ -159,7 +159,8 @@ class LeaderboardsHandler {
 	onMapLeaderboardsLoad(map: MMap) {
 		this.panels.tracksDropdown.RemoveAllOptions();
 
-		const currentMode = GameModeAPI.GetCurrentGameMode();
+		const isTabMenu = $.GetContextPanel().id === 'TabMenuLeaderboard';
```
tsa96, line 248: `this.panels.cp` instead of `$.GetContextPanel` - JS object lookup is generally faster than calling into C++. Also I would pull this out to a readonly member variable.

scripts/pages/leaderboards/leaderboards.ts
```diff
@@ -25,9 +25,13 @@ class LeaderboardsHandler {
...
-		$.RegisterForUnhandledEvent('EndOfRun_Show', (reason) => this.onShowEndOfRun(reason));
-		$.RegisterForUnhandledEvent('Leaderboards_MapDataSet', (isOfficial) => this.onMapLoad(isOfficial));
-		$.RegisterForUnhandledEvent('Leaderboards_MapLeaderboardsLoaded', (map) => this.onMapLeaderboardsLoad(map));
+		$.RegisterEventHandler('EndOfRun_Show', $.GetContextPanel(), (reason) => this.onShowEndOfRun(reason));
+		$.RegisterEventHandler('Leaderboards_MapDataSet', $.GetContextPanel(), (isOfficial) =>
+			this.onMapLoad(isOfficial)
+		);
+		$.RegisterEventHandler('Leaderboards_MapLeaderboardsLoaded', $.GetContextPanel(), (map) =>
+			this.onMapLeaderboardsLoad(map)
+		);
```
tsa96, line 34: These (and above line) should use `this.panels.cp`. Not that perf is a big deal, but it's good to be consistent.

tsa96 (changes requested):

scripts/pages/leaderboards/leaderboards.ts
tsa96: Is `button.group` even defined in your new code? It's not set in XML, and I would just use the object keys like I suggested. If you're not using the object keys, you can use `Object.values(foo => ...` instead of `Object.entries(([_, foo]) => ...`

Panzerhandschuh: Fixed. I'm not sure why but when I first used `group`, it was giving me a syntax error.

scripts/pages/leaderboards/leaderboards.ts
```diff
@@ -152,16 +181,75 @@ class LeaderboardsHandler {
...
 		this.panels.tracksDropdown.RemoveAllOptions();
 
-		const currentMode = GameModeAPI.GetCurrentGameMode();
+		const isTabMenu = this.panels.cp.id === 'TabMenuLeaderboard';
+		const currentMode = isTabMenu ? GameModeAPI.GetCurrentGameMode() : GameModeAPI.GetMetaGameMode();
 		map.leaderboards
 			.filter((leaderboard) => leaderboard.gamemode === currentMode)
+			.sort((a, b) =>
+				sortTrackId({ type: a.trackType, number: a.trackNum }, { type: b.trackType, number: b.trackNum })
+			)
```
tsa96, line 252: Having to do these `{ type: a.trackType, number: a.trackNum }` objects is *very* ugly, why do these need to be `TrackID`s anyway? I don't know when we'd need to sort a trackID like this (something that comes from timer metadata), it's much more likely we need to sort leaderboards, as in this case. Maybe `sortTrackId` (which I suggested) is a bad name, how about 
```ts
export function sortLeaderboard(
  left: { trackNum: number, trackType: TrackType },
  right:  { trackNum: number, trackType: TrackType ): number {
  return left.trackType === right.trackType ? left.trackNum - right.trackNum : left.trackType - right.trackType
} 
``` 
using an object type so any object that contains `{ trackNum: number, trackType: TrackType }` satisfies the type.

scripts/pages/leaderboards/leaderboards.ts
```diff
@@ -152,16 +173,73 @@ class LeaderboardsHandler {
...
+			return; // Load official leaderboard tracks instead
+		}
+
+		// Try to load tracks from local zones
+		const mapZoneData = MomentumTimerAPI.GetActiveZoneDefs();
+		if (mapZoneData) {
+			// Main track
+			{
+				const trackStr = $.Localize('#Leaderboards_Tracks_Main');
+				const item = $.CreatePanel('Label', this.panels.tracksDropdown, trackStr, {
```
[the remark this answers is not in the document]
tsa96: The thing about inlining `text: $.Localize('#Leaderboards_Tracks_Main')` was just because it doesn't seem helpful to define `trackStr` separate - variable name doesn't convey much information so don't think there's any point declaring it separately. But that's just a nit, not a big deal, okay as-is.

tsa96 (approved):

PR #167, opened 2024-12-28
End of run menu for new timer system

Panzerhandschuh:
Updated the end of run menu to display splits and comparisons from the new timer system.

### Checks

### DIFF
--- a/scripts/common/timer.ts
+++ b/scripts/common/timer.ts
@@ -21,23 +21,6 @@ export enum TimerState {
 	FINISHED
 }
 
-export enum ZoneType {
-	NONE = 0,
-
-	/** When majorNum = 1, and minorNum = 1 */
-	START = 1,
-
-	END = 2,
-
-	/** When majorNum > 1 and minorNum = 1 */
-	MAJOR_CHECKPOINT = 3,
-
-	/** All other timer segment zones */
-	MINOR_CHECKPOINT = 4,
-
-	CANCEL = 5
-}
-
 export interface TrackID {
 	type: uint8;
 	number: uint8;
@@ -54,15 +37,21 @@ export interface TimerStatus {
 	segmentCheckpointsCount: uint8;
 }
 
-export interface RunStats {
-	maxOverallSpeed: float;
-	maxHorizontalSpeed: float;
+export interface RunMetadata {
+	filePath: string;
+	timestamp: number;
+	gameMode: Gamemode;
+	tickInterval: float;
+	playerSteamId: number;
+	playerName: string;
+	trackId: TrackID;
+	runTime: double;
+	runSplits: RunSplits | null;
+}
 
-	overallDistanceTravelled: float;
-	horizontalDistanceTravelled: float;
-
-	jumps: uint16;
-	strafes: uint16;
+export interface RunSplits {
+	trackStats: RunStats;
+	segments: RunSegment[];
 }
 
 export interface RunSegment {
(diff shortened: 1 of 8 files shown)


styles/pages/leaderboards/leaderboards.scss
```diff
@@ -157,3 +157,7 @@
 .leaderboard-dropdown {
 	vertical-align: center;
 }
+
+.leaderboard-endofrun-button {
+	margin-left: 8px;
+}
```
tsa96, line 163: Weird naming, this should be part of a BEM block (https://getbem.com/). I would probably put this in `leaderboards` so you have `leaderboards_button`.

I know I'm being pedantic about this but keeping consistent with CSS class naming is vital, stylesheets easily become a total mess if you don't follow a shared convention.

scripts/pages/end-of-run/end-of-run.ts
```diff
@@ -522,7 +522,7 @@ class EndOfRunHandler {
 		this.panels.cp.SetDialogVariable('selected_zone', $.Localize(split.name));
 
 		// Don't scroll for the overall split
-		if (split.name !== 'Run_Comparison_Split_Overall') {
+		if (split.name !== 'Overall') {
 			this.panels.splits.FindChildTraverse(`Split${split.name}`).ScrollParentToMakePanelFit(3, false);
 		}
```
tsa96, line 548: This is a localization token; changing this will break i18n. Why change it anyway?

scripts/common/timer.ts
```diff
@@ -162,11 +161,37 @@ export function generateOverallComparisonSplit(baseRun: RunMetadata, comparisonR
...
+		generateFinishSplitComparison(
+			baseRun.runTime,
+			baseRun.runSplits,
+			comparisonRun.runTime,
+			comparisonRun.runSplits
+		)
+	];
+	comparisonSplits.push(finishSplit);
+
+	return comparisonSplits;
```
tsa96, line 194: Untested!
```ts
export function generateComparisonSplits(baseRun: RunMetadata, comparisonRun: RunMetadata): ComparisonSplit[][] {
	const comparisonSplits = baseRun.runSplits.segments.map((segment, i) =>
		segment.subsegments
			// Skip first split
			.filter((_, j) => i !== 0 && j !== 0)
			.map((_, j) => generateSegmentComparisonSplit(baseRun.runSplits, comparisonRun.runSplits, i, j))
	);

	comparisonSplits.push([
		generateFinishSplitComparison(
			baseRun.runTime,
			baseRun.runSplits,
			comparisonRun.runTime,
			comparisonRun.runSplits
		)
	]);

	return comparisonSplits;
}
```

tsa96 (changes requested):
Will look through rest of stuff tomorrow because I am sooooo sleepy

scripts/common/timer.ts
tsa96: generateSegmentComparisonSplit(baseRun.runSplits, comparisonRun.runSplits, i, i === 0 ? j - 1 : j)

Panzerhandschuh: I can fix your code by changing the filter condition to `!(i === 0 && j === 0)` and adding a `.slice(1)` after the `.map(...)`. I don't think this makes things more readable compared to the for loop though. Using a `.slice()` to account for the empty array is very unintuitive.

tsa96 (approved):
LGTM, so long as you don't want to do end of run subsegment stuff in this PR. No problem leaving it for now, but we shouldn't make the end of run issue (assuming one exists) as completely until that's done.

PR #183, opened 2025-03-01
Display accurate map information

Panzerhandschuh:
This fixes incorrect information being displayed for whether the map is linear and the zone count. This also adds bonus info to the loading screen and map selector.

### Checks

-   [ x] **I have thoroughly tested all of the code I have modified/added/removed to ensure something else did not break**
-   [ x] I have followed [semantic commit messages](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716) e.g. `feat: Add foo`, `chore: Update bar`, etc...
-   [ x] My branch has a clear history of changes that can be easy to follow when being reviewed commit-by-commit
-   [ x] My branch is functionally complete; the only changes to be done will be those potentially requested in code review
-   [ x] All changes requested in review have been `fixup`ed into my original commits.
-   [ x] Fully tokenized all my strings (no hardcoded English strings!!) and supplied [bulk JSON strings](https://docs.momentum-mod.org/guide/localization/#bulk-adding-terms) below

POEditor strings have been added already.


tsa96 (commented):
I'm very confused. Weren't we moving to using a generalized MapInfo component, that's then styled differently for its different usages? Seemed like you me and rio were all keen on that, don't understand why you're doing this this way.

Panzerhandschuh:
> I'm very confused. Weren't we moving to using a generalized MapInfo component, that's then styled differently for its different usages? Seemed like you me and rio were all keen on that, don't understand why you're doing this this way.

Most of the layouts for the map info are different. I think we should remove the following stuff:
**map-info.xml**
- Tier
- Is linear/staged
- # of stages

**tab-menu.xml**
- Is linear/staged
- # of stages

If we remove all of this, then the map info layout will be different on everything that displays that information. Also we might change the tab menu significantly so I don't think it makes sense to do a bunch of refactoring now. I tried using rio's component and it was just complicating things.

tsa96:
Right but you can do all of that using CSS like I showed you, and skip the duplicate TypeScript code. If you need me to run through that approach again I can write more when on my PC.

tsa96 (approved):

PR #199, opened 2025-04-06
Move region edit state machine to c++

PeenScreeker:
Closes momentum-mod/game/issues/2285

This pull request moves code that handles region edit state changes to c++. User facing changes:
- Unset region properties (bottom, height, safe height) are shown as blank instead of 0.00
- Deleting the last region in a zone allows the user to recreate the region for that zone. Zones are deleted from the tracklist (same as existing behavior)
- The binds list is updated to show `MOUSE2` and `ESCAPE` which function as a "back" button while editing/creating a region, and canceling the edit/creation of a region. The difference between back and cancel is subtle, cancel will always take the user straight to the menu.

Depends on red MR.

### Checks

### DIFF
--- a/layout/pages/zoning/zoning.xml
+++ b/layout/pages/zoning/zoning.xml
@@ -231,6 +231,14 @@
 						<Label class="zoning-info__bind-key" text="MOUSE1" />
 					</Panel>
 					<Panel class="zoning-info__bind">
+						<Label class="zoning-info__bind-action" text="Back" />
+						<Label class="zoning-info__bind-key" text="MOUSE2" />
+					</Panel>
+					<Panel class="zoning-info__bind">
+						<Label class="zoning-info__bind-action" text="Cancel region edit" />
+						<Label class="zoning-info__bind-key" text="ESCAPE" />
+					</Panel>
+					<Panel class="zoning-info__bind">
 						<Label class="zoning-info__bind-action" text="Delete last point" />
 						<Label class="zoning-info__bind-key" text="DELETE, BACKSPACE" />
 					</Panel>
--- a/layout/pages/zoning/zoning.xml
+++ b/layout/pages/zoning/zoning.xml
@@ -227,24 +227,26 @@
 				<Label id="SelectionMode" class="zoning__property-label" text="{s:selectionModeHint}" />
 				<Panel class="zoning-info__binds-list">
 					<Panel class="zoning-info__bind">
-						<Label class="zoning-info__bind-action" text="Select" />
+						<Label class="zoning-info__bind-action" text="{s:mouse1Hint}" />
 						<Label class="zoning-info__bind-key" text="MOUSE1" />
 					</Panel>
 					<Panel class="zoning-info__bind">
-						<Label class="zoning-info__bind-action" text="Back" />
+						<Label class="zoning-info__bind-action" text="{s:mouse2Hint}" />
 						<Label class="zoning-info__bind-key" text="MOUSE2" />
 					</Panel>
 					<Panel class="zoning-info__bind">
 						<Label class="zoning-info__bind-action" text="Cancel region edit" />
 						<Label class="zoning-info__bind-key" text="ESCAPE" />
 					</Panel>
-					<Panel class="zoning-info__bind">
-						<Label class="zoning-info__bind-action" text="Delete last point" />
-						<Label class="zoning-info__bind-key" text="DELETE, BACKSPACE" />
-					</Panel>
-					<Panel class="zoning-info__bind">
-						<Label class="zoning-info__bind-action" text="Toggle two-click mode" />
-						<Label class="zoning-info__bind-key" text="MOUSE3" />
+					<Panel id="ZoningBindsCorner" class="zoning-info__category">
+						<Panel class="zoning-info__bind">
+							<Label class="zoning-info__bind-action" text="Delete last point" />
+							<Label class="zoning-info__bind-key" text="DELETE, BACKSPACE" />
+						</Panel>
+						<Panel class="zoning-info__bind">
+							<Label class="zoning-info__bind-action" text="Toggle two-click mode" />
+							<Label class="zoning-info__bind-key" text="MOUSE3" />
+						</Panel>
 					</Panel>
 					<Panel class="zoning-info__bind">
 						<Label class="zoning-info__bind-action" text="Increase reticle distance" />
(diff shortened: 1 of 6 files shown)


scripts/pages/zoning/zoning.ts
```diff
@@ -728,110 +725,17 @@ class ZoneMenuHandler {
...
+		if (index !== -1) this.selectedZone.zone.regions[index] = newRegion;
 
+		this.showInfoPanel(false);
 		this.drawZones();
 	}
 
-	onPickCanceled() {
-		if (!this.isSelectionValid().zone || !this.selectedZone.region) return;
+	onRegionEditCanceled() {
+		if (!this.selectedZone?.region?.points.length) this.deleteRegion();
```
Panzerhandschuh, line 753: I think points is missing a ?

scripts/pages/zoning/zoning.ts
```diff
@@ -559,21 +559,37 @@ class ZoneMenuHandler {
...
-
-			selectButton?.SetSelected(true);
-			this.updateSelection(this.selectedZone.track, this.selectedZone.segment, null, deleteButton);
+			UiToolkitAPI.ShowGenericPopupTwoOptionsBgStyle(
+				$.Localize('#Zoning_Delete'),
+				$.Localize('#Zoning_LastRegion_Message'),
+				'warning-popup',
+				$.Localize('#Zoning_Delete'),
+				() => {
+					const trackIndex =
```
Panzerhandschuh, line 568: I would refactor this into a reusable function if this code is repeated more than twice.

scripts/pages/zoning/zoning.ts
```diff
@@ -559,21 +559,37 @@ class ZoneMenuHandler {
...
+					const selectButton = segmentPanel.FindChildTraverse<ToggleButton>('SelectButton');
+					const deleteButton = segmentPanel.FindChildTraverse<Button>('DeleteButton');
+
+					selectButton?.SetSelected(true);
+					this.updateSelection(this.selectedZone.track, this.selectedZone.segment, null, deleteButton);
+				},
+				$.Localize('#Zoning_Recreate'),
+				() => {
+					this.panels.zoningMenu.createRegion(
+						!this.selectedZone.segment?.checkpoints.indexOf(this.selectedZone.zone)
```
Panzerhandschuh, line 588: This should also probably use `includes()` instead of `indexOf()`

PeenScreeker: I think this is the same, we want to know if the zone is the first element in the `checkpoints` array.

I can add the elvis here, but I also kind of want to know if a segment ever exists without the `checkpoints` property (it shouldn't be possible unless a user manually edits the json file).

scripts/pages/zoning/zoning.ts
PeenScreeker: similar to the other comment, a `region` should always have a `points` property even if it's empty. Is there a better way to have an assert type error-popup behavior?

scripts/pages/zoning/zoning.ts
PeenScreeker: I can't find another repeat of this code (but it's entirely possible I'm just overlooking how this would be generalized for another similar use)

scripts/pages/zoning/zoning.ts
```diff
@@ -399,7 +395,8 @@ class ZoneMenuHandler {
 			selectButton.SetSelected(true);
 			this.updateSelection(selectionObj.track, selectionObj.segment, selectionObj.zone, deleteButton);
 			if (selectionObj.zone) {
-				this.pickCorners();
+				this.panels.zoningMenu.createRegion(!selectionObj.segment.checkpoints.indexOf(selectionObj.zone));
```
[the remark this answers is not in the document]
tsa96: `indexOf` is correct but worth being explicit with this check, makes intention clearer and requires less reasoning about JS truthiness, *especially* given it's not immediately clear that this method takes a bool param. So you want 
```js
this.panels.zoningMenu.createRegion(selectionObj.segment.checkpoints.indexOf(selectionObj.zone) === 0);
```

scripts/pages/zoning/zoning.ts
tsa96: Panzer is right about the `?.` here. In JS `?.` doesn't short-circuit, so if `this.selectedZone.segment` is `null` or `undefined`, `this.selectedZone.segment?.` will evaluated to `undefined` and throw with "trying to access properly "checkpoints" of undefined`.

And yeah the `indexOf` thing still applies, you should make the comparison explicit. So
```js
this.panels.zoningMenu.createRegion(
  this.selectedZone.segment?.checkpoints?.indexOf?.(this.selectedZone.zone) === 0
```

The elvis before the `?.indexOf` looks a bit silly, but think of the `indexOf` function as just being a property of the array (well, array prototype).

tsa96: Incidentally this is why I want to enable TypeScript strict mode, it would've flagged this as being unsafe undefined/null handling.

scripts/pages/zoning/zoning.ts
tsa96: Panzer is right, as above, ?. doesn't evaluate to like, some default initialized struct, it just creates an `undefined`. There's practically never a case where you should do `A?.B.C`, in fact TS strict mode will flag any instance of that, assuming `A` is actually possibly `undefined` or `null`.

layout/pages/zoning/zoning.xml
```diff
@@ -224,12 +224,20 @@
...
 						<Label class="zoning-info__bind-key" text="MOUSE1" />
 					</Panel>
+					<Panel class="zoning-info__bind">
+						<Label class="zoning-info__bind-action" text="Back" />
+						<Label class="zoning-info__bind-key" text="MOUSE2" />
+					</Panel>
+					<Panel class="zoning-info__bind">
+						<Label class="zoning-info__bind-action" text="Cancel region edit" />
+						<Label class="zoning-info__bind-key" text="ESCAPE" />
+					</Panel>
```
tsa96, line 240: Shouldn't this stuff be localized?

tsa96 (changes requested):

scripts/pages/zoning/zoning.ts
tsa96: Yeah unless there's genuine repeats of this, generally I avoid refactoring something out to a separate method just because it's in an arrow function. Inline arrow functions like this are very readable!

layout/pages/zoning/zoning.xml
PeenScreeker: yep, changed everything but the keybind strings.

tsa96:
Okay looking good, hope the stricter null checking saves us some hassle in the future!

tsa96 (approved):

PR #202, opened 2025-04-13
0.10.0 Comparisons, End of Run, Timer and Replay Controls

tsa96:
Several major reworks of timer-related components for 0.10.0, plus some misc things from along the way.

Sorry this is such a big PR, but I honestly didn't feel like I had fully gotten the hang of the new timer APIs until quite far into this, especially handling unordered checkpoints well.

Closes https://github.com/momentum-mod/game/issues/2292
Closes https://github.com/momentum-mod/game/issues/2291
Closes https://github.com/momentum-mod/game/issues/2287
Closes https://github.com/momentum-mod/game/issues/2222
Closes https://github.com/momentum-mod/game/issues/2275
Closes https://github.com/momentum-mod/game/issues/2209
Closes https://github.com/momentum-mod/game/issues/266
Closes https://github.com/momentum-mod/game/issues/2092

### Checks


scripts/common/web.ts
```diff
@@ -522,6 +523,33 @@ export enum GamemodeCategory {
 	DEFRAG = 8
 }
 
+export const GamemodeCategoryToGamemode = new Map<GamemodeCategory, Gamemode[]>([
```
Panzerhandschuh, line 526: Do we have any documentation telling us to modify this when adding new game modes?

scripts/common/web.ts
```diff
@@ -594,6 +622,118 @@ export enum MapSubmissionType {
...
+ *
+ * After adding tags, you need to run
+ * ```
+ * nx run scripts:maptags-poeditor
+ * ```
+ * which writes a JSON file to cwd, then import that file into POEditor, with
+ * "Also import translations to a language" set to English. POEditor will not
+ * overwrite existing translations.
+ */
+export enum MapTag {
```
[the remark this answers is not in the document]
tsa96: They have integer values we'll use on C++ in the future, also we want uniqueness.

This is a website change BTW, pulled into this repo. If you wanna review website stuff you're welcome to, but this is already merged into website

scripts/common/web.ts
tsa96: I left comments in the C++ for adding gamemodes to update the website repo yeah. These changes to panorama are entirely generated by a bash script

scripts/common/timer.ts
```diff
@@ -88,43 +96,321 @@ export interface RunSubsegment {
...
+	if (!compSplits) {
+		return [baseSeg.subsegments[baseSubsegIdx], null];
+	}
+
+	const compSeg = compSplits.segments[majorNum - 1];
+	const compSubsegIdx = compSeg.subsegments.findIndex((ss) => ss.minorNum === minorNum);
+
+	// Easy case, base and comp are at same position in splits: return respective subsegs
+	//   Base ----S1---Cp1-----Cp2----
+	//   Comp ---S1----Cp1----Cp2-----
```
Panzerhandschuh, line 169: (nit) The hyphens don't line up exactly

scripts/common/timer.ts
```diff
@@ -88,43 +96,321 @@ export interface RunSubsegment {
...
+	// Multiple segments, multiple subsegments
+	// (e.g. RJ map with courses): "X-Y" for previous majorNum, previous minorNum
+	if (segmentsCheckpointsCount > 1) {
+		// For these zones, we want:
+		//  S1----CP1----CP2----S2----CP1----CP2----END
+		// (n/a)  1-1    1-2    1-3   2-1    2-2    2-3
+		// So:
+		// Maj 1, Min 1 => N/A      Maj 1, Min 2 => 1-1      Maj 1, Min 3 => 1-2
+		// Maj 2, Min 1 => 1-3      Maj 2, Min 2 => 2-1      Maj 2, Min 3 => 2-2
+		// END => 2-3
```
Panzerhandschuh, line 250: Maybe the end zone splits should display "2-End" instead of "2-3"? Idk, I don't have a strong preference either way.

scripts/common/timer.ts
tsa96: That's deliberate, to signify the CPs being hit at different points in time.

scripts/common/timer.ts
tsa96: This is used for the name of the checkpoint you're current on, or was just completed.

If you do a surf map with two stages, no minor CPs, when you complete stage 1, the comparison will be named "1", when you complete stage 2 (and thus complete the map), it shows "2". Showing "End" instead of "2" seems very weird in that case.

Similarly if each stage had an extra minor CP, the comparisons would show "1-1", "1-2", "2-1", "2-2".

tsa96:
https://github.com/momentum-mod/panorama/pull/203 will cause a conflict. I've applied the fix from there to this branch as well, just pick changes from here when resolving.

scripts/common/timer.ts
```diff
@@ -101,43 +96,329 @@ export interface RunSubsegment {
...
+	// If this happens, something's gone wrong in the caller logic.
+	if (baseSubsegIdx === -1) {
+		throw new Error(`Missing subsegment for majorNum: ${majorNum}, minorNum: ${minorNum}`);
+	}
+
+	if (!compSplits) {
+		return [baseSeg.subsegments[baseSubsegIdx], null];
+	}
+
+	const compSeg = compSplits.segments[majorNum - 1];
```
jason-e, line 164: This shouldn't assume the comparison has segment `majorNum - 1`. I encountered this with with some comparison splits that were bugged on my end at the time, but this is something that could happen if zones differ (mostly a local zones scenario).

layout/pages/end-of-run/end-of-run.xml
```diff
@@ -5,8 +5,23 @@
...
+			<Panel class="endofrun-split diff">
+				<Panel class="endofrun-split__left">
+					<Label id="Name" class="endofrun-split__name" text="{s:name}" />
+				</Panel>
+				<Panel class="endofrun-split__right">
+					<Label class="endofrun-split__time" text="{g:time:time}" />
+					<Label class="endofrun-split__diff diff" text="{s:diff_sign}{g:time:diff}" />
+				</Panel>
+			</Panel>
+		</snippet>
```
jason-e, line 22: If there's no comparison run, these have no spacing between them 
image
vs
image

scripts/hud/comparisons.ts
```diff
@@ -1,140 +1,325 @@
...
+			this.recomputeComparisons();
+		});
+
+		$.RegisterForUnhandledEvent('OnObservedTimerCheckpointProgressed', () => {
+			const { majorNum, minorNum, segmentsCount, segmentCheckpointsCount } =
+				MomentumTimerAPI.GetObservedTimerStatus();
+
+			const splits = MomentumTimerAPI.GetObservedTimerRunSplits();
+
+			this.updateLatestSplit(
```
jason-e, line 128: I think the splits HUD needs to not do any comparison stuff if the comparison is for a different track.

In theory if your comparison run is a stage and you are running the Main track (or vice-versa) the comparisons logic could identify that and line up the comparison with the appropriate stage -- however to *actually* make that useful, what we are really going to want is the ability to do like a "auto set comparison to best of each stage" if you want to run the Main track but compare to individual stage times, so I think we will get to that later and then that would be a good time to do the stage comparison alignment thing.

scripts/hud/comparisons.ts
```diff
@@ -1,140 +1,325 @@
...
+		$.RegisterForUnhandledEvent('ComparisonRunUpdated', () => {
+			const { state } = MomentumTimerAPI.GetObservedTimerStatus();
+			// If timer just finished and we're not watching a replay, don't update - otherwise
+			// comparison will get set to the run you just finished when you PB which we
+			// obviously don't want to happen.
+			if (state === Timer.TimerState.FINISHED && this.controlledReplayID === null) {
+				return;
+			}
+
+			this.comparison = RunComparisonsAPI.GetComparisonRun();
```
jason-e, line 118: I saw a case where the splits HUD was comparing against what I'm pretty sure was an older comparison one than the one I had set (the center timer comparison HUD was showing the delta I expected to see). This might mean the game code is not calling `ComparisonRunUpdated` when it should be though so probably not a bug here. I didn't dig into it more because I was debugging something else at the time and ran out of energy/time to try to recreate it.

scripts/common/timer.ts
```diff
@@ -101,43 +96,329 @@ export interface RunSubsegment {
...
+	if (prevMaj === 0 && prevMin === 0) return split;
+
+	const [prevBase, prevComp] = findSubsegmentComparison(baseSplits, compSplits, prevMaj, prevMin) ?? [];
+
+	split.time = base.timeReached;
+	split.segmentTime = base.timeReached - prevBase.timeReached;
+
+	if (!comp) return split;
+
+	split.diff = base.timeReached - comp.timeReached;
```
jason-e, line 345: The new in-run "limited" timer splits network as 32 bit floats and then cast back to the 64 bit floats in RunSplits which introduces some extra FP error. I figure this is fine for in-run HUD stuff (vs like if you wanted to do an insanely precise comparison after the fact). However the comparison RunSplits are still the original 64 bit floats, so right now if you have a split that is the same number of ticks as the comparison, it reports a very slight difference (either plus or minus 0.00 on the HUD). This wont matter as much once we do sub-tick timing, but even then you could have truly equal splits (e.g. start from a start mark and hold forward).

So this is all to say that the comparison splits need to be cast to 32 bits then back to 64 bits before being compared to the active run splits -- if this is easy to do in the JS side that seems ideal, otherwise we could figure something out in game code. Or maybe I could just network the full doubles... 10 "limited" splits is only like 60-80 bytes the way I'm doing it, so maybe adding ~50 bytes to that would be ok (but it might be a different story if we want to add split speed too).

styles/hud/comparisons.scss
```diff
@@ -33,3 +26,74 @@
...
+	&__column {
+		flow-children: up;
+	}
+
+	// Panorama forcibly sets the `height` property when sizing text so forcing a label
+	// to a certain height is very difficult -- line-height doesn't help.
+	// This is a wrapper container that each label is placed in and vertically centered
+	// to achieve a consistent grid.
+	&__cell {
+		height: 25px;
```
jason-e, line 42: I think the rows should have a *little* more space between them, like 3-5 pixels maybe?

jason-e (changes requested):
I tested the splits/comparison stuff and the replay controls, and reviewed the comparison code but nothing else.

Aside from my few comments here, this is working REALLY well and is even working perfectly in momtv multiplayer with the RunSplits networking I just set up.

jason-e:
Also I did not test what happens with this when the limited in-run splits skip some segments/subsegments. If anything needs adjusting for that it should be minor.

scripts/hud/timer.ts
```diff
@@ -1,15 +1,14 @@
 import { PanelHandler } from 'util/module-helpers';
 import { HideHud } from 'common/state';
-import { TimerEvent_OLD, TimerState, TimerState_OLD } from 'common/timer';
+import * as Timer from 'common/timer';
 
+const DIFF_DISPLAY_TIME = 8;
```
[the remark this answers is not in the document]
tsa96: Yeah, it used to be 10. Lowered to 5.

scripts/hud/timer.ts
```diff
@@ -19,211 +18,180 @@ class HudTimerHandler {
...
+					splits,
+					this.comparison.runSplits,
+					majorNum,
+					minorNum,
+					segmentsCount,
+					segmentCheckpointsCount
+				).diff ?? 0;
+		} else if (state === Timer.TimerState.FINISHED) {
+			diff = runTime - this.comparison.runTime;
 		}
```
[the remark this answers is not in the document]
tsa96: Good spot, fixed.

scripts/common/timer.ts
tsa96: Okay, returning `null` for comparison subseg if missing

layout/pages/end-of-run/end-of-run.xml
tsa96: Fixed

styles/hud/comparisons.scss
tsa96: This okay?
Old
Momentum_Mod_-_PIX_ENABLE_-_DX11_ Development_Buil19,04,25,222615
New
Momentum_Mod_-_PIX_ENABLE_-_DX11_ Development_Buil19,04,25,222549

scripts/hud/comparisons.ts
tsa96: Yeah good spot, fixed in `438c0d5`

scripts/hud/comparisons.ts
tsa96: I think this code was wrong (below), not setting comparison to latest.
```ts
if (state === Timer.TimerState.FINISHED && this.controlledReplayID === null) {
	return;
}

this.comparison = RunComparisonsAPI.GetComparisonRun();
this.recomputeComparisons();
```
should be
```ts
this.comparison = RunComparisonsAPI.GetComparisonRun();

if (!(state === Timer.TimerState.FINISHED && this.controlledReplayID === null)) {
	this.recomputeComparisons();
}
```

tsa96: Fixed in `995469b`

scripts/common/timer.ts
tsa96: I think this is fine to do in JS with [`Math.fround`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/fround), been using that on backend a bit as well after you brought up float/double discrepancies there.

Since most of the timer comparison logic is pretty well-contained in `common/timer`, I think it's fine to just use fround everywhere in there and have docs making it clear that it's necessary to use when doing comparisons. See `415b38f` for implementation.

tsa96:
Have addressed everything you brought up. Kept the non-trivial stuff as separate commits.

styles/hud/comparisons.scss
jason-e: Rows, not columns. Though now that you've done that I think the extra column spacing is good too.

scripts/common/timer.ts
jason-e: Is that same logic used for end of run or any other places? Only in-run splits have the reduced precision, so end of run or any other places we show splits should use the full precision.

tsa96: Is there really a downside to using the lower precision? I don't think there's anywhere we show these numbers that need such a high degree of precision, and I would *really* like to avoid varying this logic between different components.

jason-e: I mean if we don't show the higher precision anywhere then we have no reason to use doubles for the timer in the first place. I figure most places would just show 2 or 3 decimal places, but then you will be able to mouse over any split/delta to see the full precision (excluding in-run splits). Even though going past ~4 decimal places might be kind of silly, I think it's cool to be able to have that in a game about timing (and it lets us flex on SourceMod communities that often only have like 2.5 places of precision with no sub-tick timing).

And speaking of conditional/varying logic, if we only use 32 bit floats for everything then we need to conditionally limit the max precision displayed to anywhere from 2 to 6 decimal places depending on the magnitude of the time.

tsa96: Sure, I hadn't realised how limited f32 accuracy is until now.

I was just writing a follow-up to my previous comment actually, I was over thinking the complexity of varying the behaviour of the common/timer.ts function. Passing some bool param is not a big deal so long as documented well. Will switch to that approach shortly, and only use the rounding for live timer.

jason-e: Also there are other ways to do this. We could add a version/parameter for GetComparisonRun that reduces the precision of all the times before returning it, and then the splits HUD would store and use that. Or maybe we should just do full precision times even in the limited splits.

tsa96: Either different C++ function or passing a bool to generateComparisons and related common/timer.ts functions sounds fine, in either case I'm most concerned having good jsdoc comments explaining when what function/bool value should be used.

Probably I'd side with doing it in JS just because it means less faffing with different red branches, but don't feel particularly strongly about either approach.

styles/hud/comparisons.scss
tsa96: Here's a handful of variants, currently we're using height: 25px, font-size: 22px. Personally I think more than a 2px increase is too much, and honestly I prefer the bottommost variation with a lower font size -- we tend to use overly large fonts IMO.

Momentum_Mod_-_PIX_ENABLE_-_DX11_ Development_Buil21,04,25,023544
Momentum_Mod_-_PIX_ENABLE_-_DX11_ Development_Buil21,04,25,023659
Momentum_Mod_-_PIX_ENABLE_-_DX11_ Development_Buil21,04,25,024013
Momentum_Mod_-_PIX_ENABLE_-_DX11_ Development_Buil21,04,25,024241
Momentum_Mod_-_PIX_ENABLE_-_DX11_ Development_Buil21,04,25,023906

Let me know what you think, my vote would be 25/21 but don't particularly mind.

scripts/common/timer.ts
jason-e: Yeah doing a bool param sounds fine especially if it's straightforward like you're saying.

styles/hud/comparisons.scss
jason-e: Can't really comment on font size because that depends on the overall resolution of the whole game, but in terms of spacing I think 28/22 looks good while 25/21 is still too compact IMO (so factoring out font size, that means the latter is 0.19 em row spacing while the former is 0.27 em, so if you wanted to do 21px font that would be about 27px height).

tsa96: Okay, going with 27/22

scripts/common/timer.ts
tsa96: Okay, done in 7861d2e

jason-e (approved):

PR #204, opened 2025-04-23
Jump timing indicator

Panzerhandschuh:
image

### Checks

### DIFF
--- a/scripts/hud/jump-timing.ts
+++ b/scripts/hud/jump-timing.ts
@@ -1,0 +1,52 @@
+import { PanelHandler } from 'util/module-helpers';
+import { RegisterHUDPanelForGamemode } from 'util/register-for-gamemodes';
+import { GamemodeCategories, GamemodeCategory } from 'common/web';
+import { RgbaTuple, rgbaTupleLerp, tupleToRgbaString } from 'util/colors';
+
+const Colors = {
+	RED: [255, 0, 0, 255] as RgbaTuple,
+	ORANGE: [255, 165, 0, 255] as RgbaTuple,
+	YELLOW: [255, 255, 0, 255] as RgbaTuple
+};
+
+@PanelHandler()
+class JumpTimingHandler {
+	readonly panels = {
+		container: $<Panel>('#JumpTimingContainer'),
+		earlyBar: $<ProgressBar>('#JumpEarlyBar'),
+		lateBar: $<ProgressBar>('#JumpLateBar'),
+		earlyLabel: $<Label>('#JumpEarlyLabel'),
+		lateLabel: $<Label>('#JumpLateLabel'),
+		earlyBarProgress: $<Panel>('#JumpEarlyBar_Left'),
+		lateBarProgress: $<Panel>('#JumpLateBar_Left')
+	};
+
+	constructor() {
+		RegisterHUDPanelForGamemode({
+			gamemodes: GamemodeCategories.get(GamemodeCategory.BHOP), // TODO: Add stamina bhop game mode
+			handledEvents: [
+				{
+					event: 'JumpTimingDataUpdate',
+					panel: this.panels.container,
+					callback: (jumpTiming, maxTimingPenalty) => this.onJumpTimingUpdate(jumpTiming, maxTimingPenalty)
+				}
+			]
+		});
+	}
+
+	onJumpTimingUpdate(jumpTiming: number, maxTimingPenalty: number) {
+		// Blend between yellow and orange for early timing, and orange and red for late timing since you lose more speed on the ground from friction and stamina
+		const earlyTimingRatio = jumpTiming < 0 ? -jumpTiming / maxTimingPenalty : 0;
+		this.panels.earlyBar.value = earlyTimingRatio;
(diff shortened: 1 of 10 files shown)


styles/hud/hud.scss
```diff
@@ -58,7 +58,6 @@ Hud,
 #HudCenter {
 	width: 100%;
 	height: 100%;
-	margin-top: 300px;
```
tsa96, line 61: Doesn't this affect other HUD components?

Panzerhandschuh: Nothing else is using that style. See [layout/hud/hud.xml](https://github.com/momentum-mod/panorama/pull/204/files#diff-18957eb64f292089f2fc9c46c1acca5a601af40648f15df8351e37ff42d63356)

scripts/hud/jump-timing.ts
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	onJumpTimingUpdate(jumpTiming: number, maxTimingPenalty: number) {
+		// Blend between yellow and orange for early timing, and orange and red for late timing since you lose more speed on the ground from friction and stamina
+		const earlyTimingRatio = jumpTiming < 0 ? -jumpTiming / maxTimingPenalty : 0;
+		this.panels.earlyBar.value = earlyTimingRatio;
+		this.panels.earlyLabel.text = jumpTiming < 0 ? -jumpTiming : 0;
+		const colorLerp = (Math.abs(jumpTiming) - 1) / (maxTimingPenalty - 1);
+		const earlyTimingColor = rgbaTupleLerp([255, 255, 0, 255], [255, 165, 0, 255], colorLerp);
+		this.panels.earlyBar.GetChild(0).style.backgroundColor = tupleToRgbaString(earlyTimingColor);
```
tsa96, line 36: I would avoid doing this call into C++ every time when you could just store this on `panels` like everything else, can also give an information name like `earlyBarProgress` or something.

tsa96 (changes requested):
Could you post a screenshot please?

Panzerhandschuh:
> Could you post a screenshot please?

I posted a screenshot on red but I'll include it here too.

styles/config.scss
```diff
@@ -692,6 +692,16 @@ $dfjump-color-background: gradient(linear, 0% 0%, 100% 0%, from(#5f5f5f80), to(#
 $dfjump-font-size: 22px !default;
 $dfjump-padding: 6px !default;
 
+// ================
+// BHOP JUMP TIMING
+// ================
+$bhopjump-width: 300px !default;
+$bhopjump-height: 20px !default;
```
tsa96, line 699: From screenshots this looks a bit tall to me, I'd go with 16px.

Couple other styling comments: 
- Labels would probably look better with at least `font-weight: medium`, maybe `semi-bold`.
- Slightly larger margin between labels and bar

Honestly, I think it should be smaller. I appreciate that it's the same width as the synchronizer bar, but that's something you constantly use to adjust your movement, whilst this is just feedback on each jump. Keep in mind we have a *lot* of possible HUD panels, more coming in the future, so the smaller the better. Almost wonder if we could align it to the right of the crosshair or something.

tsa96 (changes requested):
Sorry for reviewing this sporadically, I think this is everything now.

scripts/hud/jump-timing.ts
```diff
@@ -0,0 +1,46 @@
...
+		this.panels.earlyBar.value = earlyTimingRatio;
+		this.panels.earlyLabel.text = jumpTiming < 0 ? -jumpTiming : 0;
+		const colorLerp = (Math.abs(jumpTiming) - 1) / (maxTimingPenalty - 1);
+		const earlyTimingColor = rgbaTupleLerp([255, 255, 0, 255], [255, 165, 0, 255], colorLerp);
+		this.panels.earlyBarProgress.style.backgroundColor = tupleToRgbaString(earlyTimingColor);
+
+		const lateTimingRatio = jumpTiming > 0 ? jumpTiming / maxTimingPenalty : 0;
+		this.panels.lateBar.value = lateTimingRatio;
+		this.panels.lateLabel.text = jumpTiming > 0 ? jumpTiming : 0;
+		const lateTimingColor = rgbaTupleLerp([255, 165, 0, 255], [255, 0, 0, 255], colorLerp);
```
PeenScreeker, line 43: really minor nit, these colors could be stored in an object like this:
`const Colors = {
  YELLOW: [255, 255, 0, 255],
  ORANGE: [255, 165, 0, 255],
  RED: [255, 0, 0, 255]
}`

In this file though, you don't have a ton of colors and the comments make it pretty easy to understand what's going on. Feel free to just resolve this, just FYI.

styles/config.scss
Panzerhandschuh: Lmk if my changes are ok. I kept it the same width because it helps me see the jump timing more easily without paying as much attention to the bar. Maybe I should make this configurable?

tsa96: I guess fine for now, future styling stuff can wait, probably for same time as we do hud customizer.

tsa96 (approved):

PR #213, opened 2025-05-16
Jump timing improvements

Panzerhandschuh:
- Added support for new perf timing window
- Use events to update timing range and perf window
- Use a constant color for early timing since the speed penalty is now fixed

### Checks

### DIFF
--- a/scripts/hud/jump-timing.ts
+++ b/scripts/hud/jump-timing.ts
@@ -5,8 +5,7 @@ import { RgbaTuple, rgbaTupleLerp, tupleToRgbaString } from 'util/colors';
 
 const Colors = {
 	RED: [255, 0, 0, 255] as RgbaTuple,
-	ORANGE: [255, 165, 0, 255] as RgbaTuple,
-	YELLOW: [255, 255, 0, 255] as RgbaTuple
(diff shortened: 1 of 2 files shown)


scripts/hud/jump-timing.ts
```diff
@@ -21,31 +20,61 @@ class JumpTimingHandler {
...
+		jumpTiming = Math.max(Math.min(jumpTiming, this.range), -this.range);
+
+		if (this.bufferedJumpingEnabled) {
+			const earlyTimingRatio =
+				jumpTiming < 0
+					? Math.max(-jumpTiming - this.perfWindow, 0) / Math.max(this.range - this.perfWindow, 1)
+					: 0;
+			this.panels.earlyBar.value = earlyTimingRatio;
+			this.panels.earlyLabel.text = jumpTiming < 0 ? -jumpTiming : 0;
+			this.panels.earlyBarProgress.style.backgroundColor = tupleToRgbaString(Colors.ORANGE);
```
PeenScreeker, line 71: I don't think this is particularly expensive, so feel free to ignore to keep the early/late code looking similar. If this is constant, it could be applied once on init, or in `updateSettings()`.

PeenScreeker (approved):

PR #214, opened 2025-05-16
fix: Don't animate sticky arming independently of game state

jason-e:
This class gets applied when the stickybomb starts arming and transitions to the "armed" appearance after a set time, but this is incorrect in cases when watching a replay and then pausing or seeking. The animation was very subtle so this just removes it for now, and someone can hook up some transition styling later that is directly tied to arm time.


tsa96 (approved):
Fine with me, not a big fan of animating HUD components anyway.

PR #223, opened 2025-05-26
Safeguard hold visualizer

Panzerhandschuh:
Radial meter for visualizing held safeguarded commands.

### Checks

### DIFF
--- a/layout/hud/hud.xml
+++ b/layout/hud/hud.xml
@@ -17,6 +17,7 @@
 
 		<!-- Center of screen HUD laid out vertically top-to-bottom -->
 		<Panel id="HudCenter" hittest="false" hittestchildren="false">
+			<MomHudSafeguardIndicator />
 			<MomHudJumpTiming />
 		</Panel>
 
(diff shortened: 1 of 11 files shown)


layout/hud/hud.xml
```diff
@@ -17,6 +17,7 @@
 
 		<!-- Center of screen HUD laid out vertically top-to-bottom -->
 		<Panel id="HudCenter" hittest="false" hittestchildren="false">
+			<MomHudSafeguardIndicator />
```
jason-e, line 20: What happens if both of these HUD elements are shown? Does the safeguard indicator stay centered on the crosshair or does it move up to keep the two HUD elements centered vertically together? It seems like the point of this grouping is to dynamically position things that want to share a general area, but we really want the safeguard indicator to always be centered on the crosshair, so it should be on its own outside any of the groups.

tsa96: FWIW I think keeping it in this group for now is fine, keeps stuff aligned for now and we're going to completely remove the HUD container panels when we do the HUD customizer

PR #239, opened 2025-06-30
Zoning menu layout rework and logic improvements

jason-e:
- Change the zoning menu to be a standalone window instead of being an alternate mode of the tab menu. When `mom_zone_edit` is enabled, the zoning menu is always open rather than being tied to the tab menu bind. Instead of using the tab menu bind in combination with right click to lock the zoning menu open and interact with it, now while its open you just right click to capture the mouse to interact with it, and right or left click outside the menu to uncapture the mouse and go back to controlling the in-game camera. When the window is open but the mouse is not captured, the menu becomes translucent to not block the view as much and also make it especially clear whether the mouse is currently captured. This does make the "zone editing binds" info harder to read, but I plan to move this info to a separate small window soon to make it more obvious (and then it wont be translucent).

https://github.com/user-attachments/assets/4d781408-33d2-49cd-a30d-61459bb1ceda

- Rework the top section into three side-by-side lists roughly corresponding to tracks, segments, and zones. This is a more efficient use of space and is much easier to navigate.

image

- Streamline the internal logic for the top section to improve maintainability and minimize bugs by rebuilding the entire DOM from scratch with each change to the zones data.
- Eliminate the ability for the zones data to end up in most invalid intermediate states. For example, it is no longer possible to have a track which does not have at least one segment with at least one checkpoint (i.e. the start zone) defined. I think the only possible invalid states now are "no main track" and "no track end zone". Allowing these states is convenient for the zoning process, so the game logic will be altered to allow booting up zones in these states when local zone definitions are used.
- When adding a bonus, there are now separate buttons for "add bonus" and "add defrag modifier bonus". Once a bonus is created, it is permanently locked to that type. As such, you are no longer allowed to turn off all modifiers for a modifier bonus.
- This also incorporates global region creation and editing

### DIFF
--- a/scripts/pages/zoning/zoning.ts
+++ b/scripts/pages/zoning/zoning.ts
@@ -290,9 +290,35 @@ class ZoneMenuHandler {
 		const bonusTag = $.Localize('#Zoning_Bonus');
 		for (const [i, bonus] of this.mapZoneData.tracks.bonuses?.entries() ?? []) {
 			const selectionObj: ZoneSelection = { track: bonus, segment: null, zone: null };
-			this.addItemListItem(this.panels.leftList, `${bonusTag} ${i + 1}`, bonus, ItemType.TRACK, selectionObj);
-		}
+			const item = this.addItemListItem(
+				this.panels.leftList,
+				`${bonusTag} ${i + 1}`,
+				bonus,
+				ItemType.TRACK,
+				selectionObj
+			);
 
+			item.SetPanelEvent('oncontextmenu', () => {
+				const insertBonus = (before) => {
+					// Select this one first for the context of where to add the new one
+					this.updateSelection(selectionObj);
+					this.addBonus(before ? i : i + 1);
+				};
+
+				const contextItems = [
+					{
+						label: '#Zoning_InsertBonusBefore',
+						jsCallback: () => insertBonus(true)
+					},
+					{
+						label: '#Zoning_InsertBonusAfter',
+						jsCallback: () => insertBonus(false)
+					}
+				];
+
+				UiToolkitAPI.ShowSimpleContextMenu('', '', contextItems);
+			});
+		}
 		this.rebuildCenterList();
 		this.rebuildRightList();
 
@@ -902,12 +928,16 @@ class ZoneMenuHandler {
 		this.setInfoPanelShown(true);
 	}
 
-	addBonus() {
+	addBonus(index = null) {
 		if (!this.mapZoneData) return;
 
 		const bonus = this.createBonusTrack();
 		if (!this.mapZoneData.tracks.bonuses) {
-			this.mapZoneData.tracks.bonuses = [bonus];
+			this.mapZoneData.tracks.bonuses = [];
+		}
+
+		if (index != null && index < this.mapZoneData.tracks.bonuses.length) {
+			this.mapZoneData.tracks.bonuses.splice(Math.max(index, 0), 0, bonus);
 		} else {
 			this.mapZoneData.tracks.bonuses.push(bonus);
 		}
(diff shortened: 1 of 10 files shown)


styles/pages/zoning/zoning.scss
```diff
@@ -7,18 +7,26 @@ $medium: 28px;
...
-	padding: 4px 4px;
+	padding: 8px;
 	flow-children: down;
 	overflow: squish;
 
+	&__title {
+		color: white;
+		font-family: $font-header;
+		font-size: 40px;
+	}
```
tsa96, line 28: Oh I noticed this the other day with `&__header` -- these don't have proper font styling.

Standard way is to import `@use ../../abstract/mixin` and use `@include mixin.font-styles($use-header: true);` in this block (and `&__header`), that'll gives you correct shadows and other properties.

tsa96 (approved):

tsa96:
LGTM besides the small styling thing, but obvs Peen's review is more important.

For the TS errors you're getting,
- `Error: scripts/pages/zoning/zoning.ts(455,68): error TS2345: Argument of type 'string | GlobalRegionType' is not assignable to parameter of type 'Region | MainTrack | BonusTrack | Segment | Zone | GlobalRegionType'.` can be fixed with an `as const` like `				'#Zoning_AllowBhopZone': [GlobalRegionType.ALLOW_BHOP, 'allowBhop'] as const,`
- Other one looks like it needs `hasSelectedDefragBonus`

You could've probably figured those out but always wanna help out with the ts errors so people don't yell at me for typescript yelling at them

scripts/pages/zoning/zoning.ts
```diff
@@ -53,13 +53,29 @@ export enum RegionPolygonProblem {
...
 }
 
 interface ZoneSelection {
-	track: MainTrack | BonusTrack | null;
-	segment: Segment | null;
-	zone: Zone | null;
+	track?: MainTrack | BonusTrack | null;
+	segment?: Segment | null;
+	zone?: Zone | null;
+	global_region?: GlobalRegionSelection | null;
```
Panzerhandschuh, line 78: Why does this have an underscore? Everything else is using camelCase.

Panzerhandschuh (approved):

scripts/pages/zoning/zoning.ts
jason-e: Fixed

scripts/pages/zoning/zoning.ts
```diff
@@ -813,218 +1005,185 @@ class ZoneMenuHandler {
...
+				zone: bonus.zones.segments[0].checkpoints[0]
+			});
 
-		this.updateSelection(bonus, bonus.zones!.segments[0], bonus.zones!.segments[0].checkpoints[0], deleteButton);
+			this.panels.zoningMenu.createRegion(true);
+			this.setInfoPanelShown(true);
+		}
 	}
 
 	addSegment() {
```
PeenScreeker, line 1093: gave you snippets for the other portions of this, but left this one out. This is what I was playing with
```js
addSegment(index = null) {
	if (!this.mapZoneData) return;
	if (!this.selectedZone.track || !this.isMainTrack(this.selectedZone.track)) return;

	const newSegment = this.createSegment();

	if (!this.selectedZone.track.zones.segments) {
		this.selectedZone.track.zones.segments = [];
	}

	this.insertItem(this.selectedZone.track.zones.segments, index, newSegment);

	this.updateSelection({ track: this.selectedZone.track, segment: newSegment, zone: newSegment.checkpoints[0] });

	this.panels.zoningMenu.createRegion(true);
	this.setInfoPanelShown(true);
}
```

jason-e: I've added a context menu for inserting segments. I also added checks so each context menu is disabled if you are not allowed to create more of that thing for whatever reason.

PeenScreeker: nice, that sounds good to me

PeenScreeker (approved):

PR #243, opened 2025-07-06
Main menu ambient sounds

tsa96:
Fixes momentum-mod/game#2264

Adds ANML's new main menu sounds, complete with looping, fading, and track syncing using sound operator stacks. Has a MR on game repo with sounds and soundscript stuff.

### DIFF
--- a/scripts/pages/main-menu/main-menu-sounds.ts
+++ b/scripts/pages/main-menu/main-menu-sounds.ts
@@ -1,0 +1,90 @@
+import { Page } from './main-menu';
+import { randomInt } from 'util/functions';
+
+interface AmbientSound {
+	readonly name: string;
+	eventID: uuid | null;
+}
+
+const AmbientSounds = {
+	MainMenu: {
+		name: 'MenuAmbientMain',
+		eventID: null
+	},
+	MapSelector: {
+		name: 'MenuAmbientMapSelector',
+		eventID: null
+	},
+	Settings: {
+		name: 'MenuAmbientSettings',
+		eventID: null
+	}
+};
+
+let currentPage: Page | 'MainMenu' = 'MainMenu';
+let playedMapSelectorSparkles = false;
+
+$.RegisterForUnhandledEvent('MainMenuPageShown', (page: Page | null) => updateAmbientSounds(page ?? 'MainMenu'));
+
+$.RegisterForUnhandledEvent('GameUIStateChanged', (_oldState, newState) => {
+	// All these sounds use the "MainmenuAmbient" mixgroup, which stops annoying calls to
+	// StopAllSounds in C++ from stopping them, e.g. on level load - Panorama gets full control.
+	if (newState === GameUIState.MAINMENU) {
+		updateAmbientSounds(currentPage);
+	} else if (newState !== GameUIState.LOADINGSCREEN) {
+		stopAmbientSounds();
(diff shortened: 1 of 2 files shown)


scripts/util/functions.ts
```diff
@@ -55,3 +55,11 @@ export function* traverseChildren(panel: GenericPanel): Generator<GenericPanel>
...
 	}
 }
+
+export function randomInt(min: number, max: number): number {
+	if (min > max) {
+		[max, min] = [min, max];
+	}
+
+	return Math.floor(Math.random() * (max - min + 1) + min);
+}
```
Gocnak, line 65: Is this inclusive of min and max? `randomInt(2, 4)`  could be over 4 if `Math.random()` returns 0.9999. Is the `+1` needed?

scripts/pages/main-menu/main-menu-sounds.ts
```diff
@@ -0,0 +1,90 @@
...
+		eventID: null
+	}
+};
+
+let currentPage: Page | 'MainMenu' = 'MainMenu';
+let playedMapSelectorSparkles = false;
+
+$.RegisterForUnhandledEvent('MainMenuPageShown', (page: Page | null) => updateAmbientSounds(page ?? 'MainMenu'));
+
+$.RegisterForUnhandledEvent('GameUIStateChanged', (_oldState, newState) => {
```
Gocnak, line 29: Does this not need to use a constructor?

scripts/util/functions.ts
tsa96: We're flooring so can't be over 4, the possible return values for this are 2, 3 and 4. I think it's more intuitive for a such a function to range over a closed interval than half-open -- `randomInt(2, 4)` returning either 2 or 3 seems odd to me.

scripts/pages/main-menu/main-menu-sounds.ts
tsa96: Technically no, but it's complicated.

In the past, JS classes where mostly useful as a way to ensure that multiple scripts import from XML have separate context (which is essentially the same as scope). If you `<include>ed` two scripts and both defined `foo`, the second script would error. To solve you can either wrap everything in an IIFE that returns an object, which CS:GO did, or use a class, which we used to do (with how JS classes work under the hood, those two are very similar things anyway).

JS modules automatically have their own context and scope, you can run two modules inside an `<include>` block and they'll be completely independent, though you could still import one from the other in JS and C++ will resolve them. Stuff defined in those modules are *not* immediately exposed to the main panel JS context that panel event scripts run in, to expose them you need to fetch that main JS context object from C++ with `$.GetContextObject()` and set properties on that object. Sorry, I'm glossing over a lot of complicated JS stuff here, but in JS, context is an object. Since we're using TypeScript now, we handle exposing classes to the main JS context object using `@PanelHandler`, which also automatically calls the class's ctor as well (Honestly it's a bit weird, but nice to use).

So for something like `MainMenuHandler` where buttons and stuff in XML want to register events that call into JS code, it makes sense to wrap in a class and expose. But for this file, all we're doing is registering some events and calling C++ functions when they're fired. It'd be easy to turn this into a class if needed, but honestly for really small scripts like this I prefer keeping it super simple.

Gocnak (approved):

PR #272, opened 2025-08-06
Fix main menu ambience in maps

tsa96:
Closes momentum-mod/game/issues/2412

GameUI events are evil and fire all over the fucking place so need bunch of extra checks and faff

### Checks

        break**
        review
-   [x] Fully tokenized all my strings (no hardcoded English strings!!) and supplied


Panzerhandschuh:
Would it be better to control the ambient audio from C++?

tsa96:
> Would it be better to control the ambient audio from C++?

No, we'd be hooking the exact same events so code would be practically the same. I would always rather do small scripts like this in JS than C++.

Gocnak (approved):
Your comments about game UI state had me looking through the code and the stuff Valve has for CS:GO menu music is a complete mess. We should definitely clean that up and probably do this properly from C++ after 0.10.

PR #274, opened 2025-08-11
Update Roboto fonts to variable versions

EnnuiL:
This updates the Roboto fonts used on Momentum Mod's GUI to their newer versions, where they are now variable fonts. While unfortunately Panorama itself can't really take advantage of it through numerical `font-weight`s and `font-variant`s, the font libraries used by the engine can simply use them and everything will run normally.

The specific fonts used were [Roboto v3.0.11](https://github.com/googlefonts/roboto-3-classic/releases/tag/v3.011) and [the latest Roboto Mono on Google Fonts](https://fonts.google.com/specimen/Roboto+Mono). ~~In the case of Roboto v3.0.11, the web version was used since as far as I know, the Condensed version is inaccessible from Panorama, and the web version splits it into a separate font.~~ Currently, the complete hinted version is being used, this will need some discussion I guess. 

I took screenshots and compared the fonts versions on the main menu, and my conclusion is that the difference is pretty much invisible (at least on 1440p)

### Checks

        break**
        review
-   [ ] ~~Fully tokenized all my strings (no hardcoded English strings!!) and supplied
        [bulk JSON strings](https://docs.momentum-mod.org/guide/localization/#bulk-adding-terms) below~~


EnnuiL:
Ok, I double-checked and can confirm that Panorama currently doesn't let you change the font width, therefore using the web version is probably for the best

Gocnak:
Can we clean up the names of the `.ttf` files? `Roboto[ital,wght]` is really gross. Also, is Pano even using these new files, or are you missing a change to consume the new files added here?

EnnuiL:
> Can we clean up the names of the `.ttf` files? `Roboto[ital,wght]` is really gross. Also, is Pano even using these new files, or are you missing a change to consume the new files added here?

Ah sure, I can do that; Also yes, Panorama is using the new files because if you nuke the original Roboto TTFs, it will fall back to the fallback font (which I'm not sure what it is), but once added the new variable TTF, Pano will use it just fine, it's just that the differences aren't that noticeable without comparing pixels. This version however, should technically have better language support, but I haven't been able to put it to test.

tsa96 (approved):

PR #275, opened 2025-08-11
Improve non-Latin headers through Bebas Neue Momentum and `text-transform: uppercase`

EnnuiL:
Closes momentum-mod/game/issues/1947
Closes momentum-mod/game/issues/2010

This uses the improved [Bebas Neue Momentum](https://github.com/momentum-mod/styling/tree/main/fonts/BebasNeue) font that the website already uses in order to ensure that more characters are covered. Notably, this fixes Cyrillic headers.

Also, this adds `text-transform: uppercase` in order to ensure that only capital letters are used, instead of relying on the font being all uppercase.

PS: Since Bebas Neue Momentum is being used, there may be slight differences such as slashed zeros (addressing an issue) and longer underlines (which might be bad?).

| Old | New |
|--------|--------|
| A screenshot of Momentum Mod's settings page in Russian as of 0.10.0-pre40 | A screenshot of Momentum Mod's settings page in Russian, with the PR applied |

### Checks

        break**
        review
-   [ ] ~~Fully tokenized all my strings (no hardcoded English strings!!) and supplied
        [bulk JSON strings](https://docs.momentum-mod.org/guide/localization/#bulk-adding-terms) below~~


styles/pages/settings/settings.scss
```diff
@@ -260,7 +261,7 @@ $page-width: 1920px - $rightnav-width;
 	&__subitem-label {
 		padding-left: 24px;
 
-		@include mixin.font-styles($use-header: true, $light-header: true);
```
Gocnak, line 263: There are instances like these where both were true passed in, but the separation of the mixins prevents `font-family` from ever being the "light" header version. Is this intentional?

PR #278, opened 2025-08-13
Run style dropdown

### LINKED ISSUE #2149: Leaderboard Style Selection
Style selection UI on the leaderboard / tab menu
Image

Panzerhandschuh:
Closes [#2149](https://github.com/momentum-mod/game/issues/2149)

I'm opening this up for early review, but we don't need to merge this until after 0.10 releases.

momentum_MGF0ie7Nkt

### Checks

        break**
-   [ ] My branch is functionally complete; the only changes to be done will be those potentially requested in code
        review
-   [ ] Fully tokenized all my strings (no hardcoded English strings!!) and supplied


layout/pages/leaderboards/leaderboards.xml
```diff
@@ -45,6 +45,10 @@
 
 			<Panel class="w-fill mr-4"/>
 
+			<DropDown id="StylesDropdown" class="dropdown leaderboards__style-dropdown" menuclass="dropdown-menu">
+				<Panel id="placeholder" /> <!-- This is needed to unfuck the dynamic dropdown -->
```
jason-e: I know this comment was copied from below but it seems unprofessional and could probably do a better job of explaining what this is fixing (if you know).

scripts/pages/leaderboards/leaderboards.ts
```diff
@@ -347,8 +348,17 @@ class LeaderboardsHandler {
 			this.panels.stylesDropdown.AddOption(item);
 		});
 
+		const defaultStyle = GamemodeDefaultStyle.get(currentMode) ?? Style.NORMAL;
```
jason-e: This information should come from the game code, not the website definitions. We should really avoid doing this thing as much as possible where the website types are dictating how the game works when the same information is available from C++.

There's also a bigger problem here where this dropdown is trying to pick what to default to on it's own while at the same time the C++ is choosing how to populate the window by default. One should be based on the other, where say the C++ picks what to display by default, and then this code checks what the leaderboard is showing at the time of init and selects that in the dropdown. This would also avoid this logic even needing to understand "gamemode defaults" at all, since it would just defer to whatever C++ did.

tsa96 (approved):
LGTM, might change styling a bit in the future but not high priority.

jason-e (approved):

Gocnak (approved):

PR #286, opened 2025-08-27
Make time formatting identical to MomUtil::FormatTime

tsa96:
Could expose the C++ function directly if we wanted, doing this because it's easier/I'm lazy. We need to have an identical TypeScript implementation to C++ anyway since dashboard uses it.

Same thing as https://github.com/momentum-mod/website/pull/1287


tsa96:
Never mind, this is stupid, I'll hook it up to C++.

TEKN1SHUN (approved):

PR #290, opened 2025-09-09
Maintain scroll position after updating selection

TEKN1SHUN:
Closes momentum-mod/game/issues/2452

This pull request implements the following changes:
- keep scroll position of selection lists when new zoning object is selected
- include global regions when counting total regions (c++ validator also uses this behavior)
- don't require the user to submit changes to test entry fields
- ~~teleport to region when new zoning object is selected (don't require double click)~~ discarded change
- add a new button for previewing a teleport destination
- when teleporting to view a region, use the improved teleport (don't send the user to the TP destination)
- when the user teleports to a defrag bonus, use the main track's zones

Requires red MR for teleport features to be merged.

### Checks

        break**
        review
-   [x] Fully tokenized all my strings (no hardcoded English strings!!) and supplied

### DIFF
--- a/scripts/pages/zoning/zoning.ts
+++ b/scripts/pages/zoning/zoning.ts
@@ -261,7 +261,11 @@ class ZoneMenuHandler {
 			} else if (selectionWhenActivated.segment) {
 				region = selectionWhenActivated.segment.checkpoints[0].regions[0];
 			} else if (selectionWhenActivated.track) {
-				region = selectionWhenActivated.track.zones.segments[0].checkpoints[0].regions[0];
+				if (this.isDefragBonus(selectionWhenActivated.track)) {
+					region = this.mapZoneData.tracks?.main.zones.segments[0].checkpoints[0].regions[0];
+				} else {
+					region = selectionWhenActivated.track.zones.segments[0].checkpoints[0].regions[0];
+				}
 			} else if (selectionWhenActivated.globalRegion?.index >= 0) {
 				region = this.selectedZone.globalRegion.regions[this.selectedZone.globalRegion.index];
 			}
--- a/scripts/pages/zoning/zoning.ts
+++ b/scripts/pages/zoning/zoning.ts
@@ -1005,6 +1005,12 @@ class ZoneMenuHandler {
 		this.panels.zoningMenu.moveToRegion(this.selectedRegion);
 	}
 
+	previewTeleDest() {
+		if (!this.selectedZone || !this.selectedRegion) return;
+
+		this.panels.zoningMenu.previewTeleDest(this.selectedRegion);
+	}
+
 	pickCorners() {
 		if (!this.selectedZone || !this.selectedRegion) return;
 		this.panels.zoningMenu.editRegion(PickType.CORNER);
--- a/scripts/pages/zoning/zoning.ts
+++ b/scripts/pages/zoning/zoning.ts
@@ -1022,7 +1022,6 @@ class ZoneMenuHandler {
 
 		const height = Number.parseFloat(this.panels.regionHeight.text);
 		this.selectedRegion.height = Math.max(Number.isNaN(height) ? 0 : height, 1);
-		this.panels.regionHeight.text = this.selectedRegion.height.toFixed(2);
 
 		this.updateEditorRegions();
 	}
--- a/scripts/pages/zoning/zoning.ts
+++ b/scripts/pages/zoning/zoning.ts
@@ -1628,6 +1628,9 @@ class ZoneMenuHandler {
 		for (const bonus of this.mapZoneData.tracks.bonuses ?? []) {
 			regionCount += countTrackRegions(bonus);
 		}
+		for (const [regions] of Object.values(this.mapZoneData.globalRegions ?? {})) {
+			regionCount += regions?.length ?? 0;
+		}
 
 		return regionCount;
 	}
--- a/scripts/pages/zoning/zoning.ts
+++ b/scripts/pages/zoning/zoning.ts
@@ -104,9 +104,11 @@ class ZoneMenuHandler {
 	readonly panels = {
 		zoningMenu: $.GetContextPanel<ZoneMenu>()!,
 
-		leftList: $<Panel>('#LeftList')!,
-		centerList: $<Panel>('#CenterList')!,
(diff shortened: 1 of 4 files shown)


scripts/pages/zoning/zoning.ts
```diff
@@ -334,15 +336,23 @@ class ZoneMenuHandler {
...
-		this.panels.leftList.RemoveAndDeleteChildren();
-		this.panels.centerList.RemoveAndDeleteChildren();
-		this.panels.rightList.RemoveAndDeleteChildren();
+		const oldScrollOffset = {};
+
+		// Record the current scroll position. If the panel is overflowing,
+		// this will be used to match the scroll position after updating list panels
+		for (const [position, panel] of Object.entries(this.panels.selectionLists)) {
+			oldScrollOffset[position] = panel?.GetParent()?.scrolloffset_y ?? 0;
+		}
```
jason-e, line 345: Would be better to use a name like "listIndex" instead of "position" since "position" sounds like it's the scroll position.

scripts/pages/zoning/zoning.ts
```diff
@@ -444,10 +454,26 @@ class ZoneMenuHandler {
 				this.selectedZone.globalRegion?.type == null ||
 				this.selectedZone.globalRegion.type === GlobalRegionType.TYPES_LIST
 		);
+
+		// Restore scroll position recorded before lists were rebuilt
+		for (const [position, panel] of Object.entries(this.panels.selectionLists)) {
```
jason-e, line 459: "listIndex" instead of "position" here as well

scripts/pages/zoning/zoning.ts
```diff
@@ -444,10 +454,26 @@ class ZoneMenuHandler {
...
+			const scrollParent = panel?.GetParent();
```
jason-e, line 460: AFAIK a panel always has a parent, so this check seems unnecessary.

jason-e (changes requested):
This doesn't seem to be working for the leftmost list. Selecting anything in any list seems to always cause the left list to scroll to the bottom.

scripts/pages/zoning/zoning.ts
tsa96: Seems like can safely remove that check and `?.`, aren't the selection lists non-null anywhere?

Type definition for `GetParent` has it possibly returning null since technically the outermost panels of base layouts don't have parents. Could update the definition to not include that and leave a note in jsdoc, don't really mind. Otherwise would need to do `GetParent()!` to make TS happy in strict-mode.

scripts/pages/zoning/zoning.ts
tsa96: I would use Object.values and use the panel IDs as keys. I tend to avoid using numeric keys in Objects in general - it works okay in this case, but the index keys are stringified when inserting and accessing, whilst giving the impression of being numbers, can be a bit hard to understand.

scripts/pages/zoning/zoning.ts
```diff
@@ -1584,6 +1630,9 @@ class ZoneMenuHandler {
 		for (const bonus of this.mapZoneData.tracks.bonuses ?? []) {
 			regionCount += countTrackRegions(bonus);
 		}
+		for (const [_, regions] of Object.entries(this.mapZoneData.globalRegions) ?? []) {
+			regionCount += regions?.length ?? 0;
+		}
```
tsa96, line 1635: May as well be `Object.values`

tsa96 (changes requested):

scripts/pages/zoning/zoning.ts
TEKN1SHUN: @jason-e does it make sense to just call the key "list" since they're strings rather than numbers ('left', 'center', and 'right')?

tsa96: Yeah "list" seems fine

tsa96 (approved):

scripts/pages/zoning/zoning.ts
jason-e: Sorry to nitpick again but list is also weird because it sounds like its an array and not a string. Maybe "listKey", or list "listType" etc

scripts/pages/zoning/zoning.ts
```diff
@@ -1630,6 +1630,9 @@ class ZoneMenuHandler {
 		for (const bonus of this.mapZoneData.tracks.bonuses ?? []) {
 			regionCount += countTrackRegions(bonus);
 		}
+		for (const [regions] of Object.values(this.mapZoneData.globalRegions) ?? []) {
```
jason-e, line 1638: This needs to be:

```js
for (const [regions] of Object.values(this.mapZoneData.globalRegions ?? {})) {
```

jason-e:
I'm not sure it's good to *always* teleport on left click with no other option. For example I selected a segment to look at the settings, got teleported to the region start, moved a little bit to get a different vantage point, then wanted to check the zone settings and clicked the zone, and get teleported to it again (losing my preferred nearby vantage point) even though I was really already there.

What if instead of teleport on left click, it was something like teleport on ctrl+click? Teleporting would be optional but would be a bit more convenient than double clicking. Haven't tried it but might be better. Open to other suggestions.

TEKN1SHUN:
> I'm not sure it's good to _always_ teleport on left click with no other option. For example I selected a segment to look at the settings, got teleported to the region start, moved a little bit to get a different vantage point, then wanted to check the zone settings and clicked the zone, and get teleported to it again (losing my preferred nearby vantage point) even though I was really already there.
> 
> What if instead of teleport on left click, it was something like teleport on ctrl+click? Teleporting would be optional but would be a bit more convenient than double clicking. Haven't tried it but might be better. Open to other suggestions.

I haven't looked into it so maybe not as big a deal as I'm thinking. I'm not sure how to implement modifier keys for pano events. It's really easy in c++, haha, maybe the answer is move even more to c++.

I think double click is still the best, but I've also grown used to teleporting on single click. I wonder if it makes sense to treat the left and center lists differently than the right list?

Another alternative would be to store the last teleported region, and only move to a new location if you are moving to a new region. This would prevent re-teleporting when different list entries actually point to the same region.

jason-e (approved):

PR #295, opened 2025-09-17
Fix obvious text clipping issues with other languages

EnnuiL:
This addresses the most obvious issues involving non-English languages, including:

- The "News" title being cutoff
- The settings headers being cutoff
- The dropdown offsetting the label vertically down

Obvious being defined as "I can see this within 5 minutes"

More information (including a comparison table) and more testing tomorrow

### Checks

        break**
        review

### DIFF
--- a/styles/pages/main-menu.scss
+++ b/styles/pages/main-menu.scss
@@ -499,7 +499,6 @@ $drawer-width: 768px;
 
 	&__header {
 		width: 100%;
-		height: 31px;
 		background-color: rgba(0, 0, 0, 0.4);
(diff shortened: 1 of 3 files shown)


styles/pages/main-menu.scss
```diff
@@ -499,7 +499,6 @@ $drawer-width: 768px;
 
 	&__header {
 		width: 100%;
-		height: 31px;
```
Gocnak, line 502: Does outright removing the height here make sense versus just increasing it? What do these changes look like?

Gocnak: Thanks! Looks good to me, we'll pull this in for the next release.

PR #303, opened 2025-09-28
Point localization folder to proper folder

Gocnak:
We use resource for it because VGUI uses some of the strings still too iirc?


tsa96:
Yeah don't really care where they're stored, lgtm

PR #347, opened 2026-03-26
Fix timer status JS exception entering cancel zones

tsa96:
Closes momentum-mod/game/issues/2661

Fix for  a rare case of leaving a start zone and entering a cancel zone on same tick e.g. here:
Momentum_Mod_-_DX1126,03,26,075057

There's maybe some other cases where a similar error could happen in `common/timer.ts` but I'm not sure and honestly don't have the brainpower at the moment to dive into the logic there.

### Checks

-   [x] **I have thoroughly tested all of the code I have modified/added/removed to ensure something else did not
        break**
-   [x] I have followed [semantic commit messages](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716)
        e.g. `feat: Add foo`, `chore: Update bar`, etc...
-   [x] My branch has a clear history of changes that can be easy to follow when being reviewed commit-by-commit
-   [x] My branch is functionally complete; the only changes to be done will be those potentially requested in code
        review
-   [x] All changes requested in review have been `fixup`ed into my original commits.
-   [x] Fully tokenized all my strings (no hardcoded English strings!!) and supplied
        [bulk JSON strings](https://docs.momentum-mod.org/guide/localization/#bulk-adding-terms) below

PR #348, opened 2026-03-31
fix: implement fallback logic

YelehaUwU:
This PR fixes an issue where crossing from the start to the end without any splits on some maps, where this is possible, would cause a JS exception.

### Checks

-   [X] **I have thoroughly tested all of the code I have modified/added/removed to ensure something else did not
        break**
-   [X] I have followed [semantic commit messages](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716)
-   [X] My branch has a clear history of changes that can be easy to follow when being reviewed commit-by-commit
-   [X] My branch is functionally complete; the only changes to be done will be those potentially requested in code
        review
-   [X] All changes requested in review have been `fixup`ed into my original commits.
-   [X] Fully tokenized all my strings (no hardcoded English strings!!) and supplied

### DIFF
--- a/scripts/hud/status.ts
+++ b/scripts/hud/status.ts
@@ -134,8 +134,19 @@ class HudStatusHandler {
 
 		// Use the subsegment count as the checkpoint number if checkpoints aren't ordered
 		const splits = MomentumTimerAPI.GetObservedTimerRunSplits();
-		const segment = splits.segments[majorNum - 1];
-		const checkpointNum = segment.checkpointsOrdered ? minorNum : segment.subsegments.length;
+		let segment, checkpointNum;
+		if (
+		    splits &&
+		    Array.isArray(splits.segments) &&
+		    splits.segments.length >= majorNum &&
+		    splits.segments[majorNum - 1]
+		) {
+		    segment = splits.segments[majorNum - 1];
+		    checkpointNum = segment.checkpointsOrdered ? minorNum : segment.subsegments.length;
+		} else {
+		    segment = null;
+		    checkpointNum = minorNum || 1;
+		}
 
 		// state is TimerState.RUNNING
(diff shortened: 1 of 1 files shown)


tsa96 (changes requested):
Can you explain exactly what bug you're encountering? In https://github.com/momentum-mod/panorama/pull/347 I'm addressing essentially the same thing, but in that case it's addressing a known timer complication we decided to just handle in JS (haven't updated that PR since conversation with rio sorry)

tsa96:
> Can you explain exactly what bug you're encountering? In https://github.com/momentum-mod/panorama/pull/347 I'm addressing essentially the same thing, but in that case it's addressing a known timer complication we decided to just handle in JS (haven't updated that PR since conversation with rio sorry) 

It sounds very similar to what we were encountering. Could you give an example map and let me know the exact exception that's firing? Also I think since this is a function designed to return a string that happens to misbehave for a single frame, for code simplicity the best practice is to early return an empty string in that instance, like my PR is doing. Feel free to discuss here or #panorama in discord though!

YelehaUwU:
You can try it out on the map kitsune by going from white zone to red zone and back. You get a js exception pop up. This pr just makes sure it won’t error out.

tsa96:
Okay yeah, it's the exact same issue as https://github.com/momentum-mod/panorama/pull/347. rio's taken a look at the exact case since I made that PR, so I've updated that to detail what's causing the bug a bit more in a comment. It's complicated C++ behaviour so worth making clear what's going on.

Our changes are functionally equivalent now, so I'm going to go with mine since it includes the explanation of what's happening. Appreciate you looking into this though, have added you as a co-author to the commit in case you care about the Github reputation :)

PR #351, opened 2026-04-30
HUD Customizer

tsa96:
Opening a draft PR for this so I can review Natan's changes.

Since this PR has *so* many commits we should probably squash rather than rebase.

### Checks

        break**
-   [ ] My branch is functionally complete; the only changes to be done will be those potentially requested in code
        review
-   [ ] Fully tokenized all my strings (no hardcoded English strings!!) and supplied


scripts/hud/customizer.ts
```diff
@@ -0,0 +1,1591 @@
...
+				case CustomizerPropertyType.FONT_PICKER: {
+					panel.LoadLayoutSnippet('dynamic-fontpicker');
+					panel.SetDialogVariable('name', dynamicStyle.properties.name);
+					const dropdown = panel.FindChildTraverse<DropDown>('DropDown')!;
+
+					for (const font of this.fonts) {
+						const panel = $.CreatePanel('Label', dropdown, font);
+						panel.text = font;
+						// Disabled because it causes extreme lag on linux ( can't test windows )
+						// panel.style.fontFamily = font;
```
tsa96: How bad is it? How many fonts do you have installed on the system? Quite annoying not to have this.

scripts/hud/ammo.ts
```diff
@@ -19,6 +19,12 @@ class MomHudAmmoHandler {
 		registerHUDCustomizerComponent($.GetContextPanel(), {
 			resizeX: true,
 			resizeY: false,
+			gamemode: [
+				Gamemode.RJ,
+				Gamemode.SJ,
+				Gamemode.CONC,
```
tsa96: Doesn't `GamemodeCategoryToGamemode` has entries for RJ/SJ/Conc? Should use those, even if it's unlikely we ever add more modes to those categories. Should probs follow that rule in most cases, use gm categories not gms, unless for stuff genuinely specific to a certain gamemode.

scripts/hud/customizer.ts
```diff
@@ -0,0 +1,1591 @@
...
+	private _enabled: boolean = undefined!;
+	private _offsetX: number = undefined!;
+	private _offsetY: number = undefined!;
+	private _width: number | undefined = undefined!;
+	private _height: number | undefined = undefined!;
+
+	private static userLayout: HudLayout | undefined;
+	private static defaultLayout: HudLayout;
+
+	//A record of CustomizerSettings dynamic styles <StyleID, value>
```
tsa96: Huge nit, but please put a space between `//` and your comments. We're not being particular about commit history with this PR (we'll just squash), should be able to fix with a regex over the whole project. `\/\/(?=\S)` should work.

scripts/hud/customizer.ts
```diff
@@ -0,0 +1,1591 @@
...
+		// components with special handling know to prepare for editing mode.
+		$.RegisterForUnhandledEvent('HudCustomizer_OpenedInternal', () => this.enableEditing());
+		$.RegisterForUnhandledEvent('HudCustomizer_Closed', () => this.disableEditing());
+
+		// TODO: Below todo is *probably* fine now using events like this, but be very careful everything
+		// is getting registered okay.
+		// TODO (Old): I *think* we're gonna need this event to cover all cases where the HUD is reloaded.
+		// Looks like some stuff like HudSpecInfoHandler listening for PanelLoaded is getting
+		// called later than this though...
+		$.RegisterForUnhandledEvent('MapCache_MapLoad' as any, () => {
```
tsa96, line 609: Oh, if this event isn't in pano-typed events.d.ts (think that's what it's called, you should add it).

scripts/hud/customizer.ts
```diff
@@ -88,7 +88,7 @@ const OVERLAY_MARGIN = 32;
 
 interface DynamicStyle {
 	properties: Readonly<DynamicStyleProperties>;
-	value: any;
+	value?: any;
```
tsa96, line 98: I suppose whether `value: any` and `value?: any` express different types is a question for the philosophers... Guess it has some explanatory use here regardless

scripts/hud/customizer.ts
```diff
@@ -839,6 +869,51 @@ class HudCustomizerHandler implements IHudCustomizerHandler {
 					this.updateActiveComponentOverlayPosition();
 				}
 			}
+		};
+
+		const childrenStyleIDs = new Set<StyleID>();
+
+		for (const [_, dynamicStyle] of Object.entries(component.dynamicStyles)) {
```
tsa96: Use Object.values

tsa96: Or if you want to be silly you could 
```js
const childrenStyleIDs = new Set(
  Object.values(component.dynamicStyles)
    .filter(({ children }) => children)
    .flatMap(({ children }) => (Array.isArray(children) ? children : [children]))
);
```
(don't need to change to this, I'm just bored)

scripts/hud/customizer.ts
```diff
@@ -839,6 +869,51 @@ class HudCustomizerHandler implements IHudCustomizerHandler {
...
+
+		const appendChildren = (parentStyleID: StyleID, parentWrapper: Panel) => {
+			const parentStyle = component.dynamicStyles[parentStyleID];
+			const children = parentStyle.properties.children;
+			if (children) {
+				const normalizedChildren = Array.isArray(children) ? children : [children];
+
+				const childrenWrapper = $.CreatePanel('Panel', parentWrapper, 'ChildrenWrapper');
+				childrenWrapper.AddClass('hud-customizer-settings__row-wrapper--children');
+				childrenWrapper.AddClass('hud-customizer-settings__row-wrapper--children-hidden');
```
tsa96: Was getting confused by naming here, could you do `--has-children` or something to make clear this is a modifier. At first it seemed like you were just getting the BEM semantics wrong.

Also, you can just do 
```ts
$.CreatePanel(
  'Panel',
  parentWrapper,
  'ChildrenWrapper',
  { class: 'hud-customizer-settings__row-wrapper__children hud-customizer-settings__row-wrapper--children-hidden' }
)
```
I guess with Prettier on it'll end up as more lines, but still prefer that to calling into C++ unnecessarily.

scripts/hud/customizer.ts
```diff
@@ -839,6 +869,51 @@ class HudCustomizerHandler implements IHudCustomizerHandler {
...
+
+				for (const child of normalizedChildren) {
+					const childStyle = component.dynamicStyles[child.styleID];
+					if (!childStyle) continue;
+
+					createStylePanel(child.styleID, childStyle, childrenWrapper);
+
+					if (childStyle.properties.children) appendChildren(child.styleID, childrenWrapper);
```
tsa96, line 1715: This is just a thing to randomly point out, personally I prefer to use `{}` braces around if statements like this to force Prettier to wrap them -- find it a bit harder to read without wrapping. An annoying feature of Prettier (which by design isn't configurable). Generally these days I use braces for everything besides `return`, `continue` control flow etc... though I used to be less consistent. Not a big deal, don't need to go through and change anything, would just recommend for future

scripts/hud/customizer.ts
```diff
@@ -839,6 +869,51 @@ class HudCustomizerHandler implements IHudCustomizerHandler {
...
+				}
+			}
+		};
+
+		// Generate settings based on CustomizerPropertyType
+		for (const [styleID, dynamicStyle] of Object.entries(component.dynamicStyles)) {
+			if (childrenStyleIDs.has(styleID)) continue;
+
+			const wrapper = $.CreatePanel('Panel', this.panels.activeComponentSettingsList, 'ParentWrapper');
+			wrapper.AddClass('hud-customizer-settings__row-wrapper');
```
tsa96: Another `AddClass` call that could be in `CreatePanel`

scripts/hud/strafe-trainer.ts
```diff
@@ -210,20 +263,111 @@ class StrafeTrainer {
...
+					expandable: true,
+					children: [{ styleID: 'backgroundColor' }, { styleID: 'needleColor' }, { styleID: 'colorByGain' }]
+				},
+				//TODO: Add highlight color ( this.altColor );
 				backgroundColor: {
 					name: 'Background Color',
 					type: CustomizerPropertyType.COLOR_PICKER,
+					targetPanel: ['.strafetrainer__background', '.strafetrainer__container'],
+					callbackFunc: (panel, value) => {
+						const splitRGBA = this.splitRgbFromAlpha(value);
```
tsa96: Weird function just for this, can't you just use `rgbaStringToTuple(value)[4]`?

scripts/common/hud-customizer.ts
```diff
@@ -14,6 +14,7 @@ export function registerHUDCustomizerComponent(panel: GenericPanel, properties:
 }
 
 export enum CustomizerPropertyType {
+	NONE,
```
tsa96, line 17: This is an odd name for this, wouldn't something like `CONTAINER` be a lot clearer? Maybe I've missed something, but this gave me the impression it was for something more generic than just grouping children.

scripts/hud/status.ts
```diff
@@ -72,49 +73,49 @@ class HudStatusHandler {
...
+				},
 				fontColor: {
 					name: 'Font Color',
 					type: CustomizerPropertyType.COLOR_PICKER,
 					targetPanel: '.hudstatus__label',
-					styleProperty: 'color'
+					styleProperty: 'color',
+					callbackFunc: (panel, value) => {
+						panel.style.textShadowFast = this.getAdjustedTextShadow(value as rgbaColor);
+					}
```
tsa96: Really not crazy about this approach tbh - wouldn't the offset of the shadow want to be proportional to the size of the font in some way? I assumed we'd make a `CustomizerPropertyType.SHADOW_PICKER` at some point. Or if that's too wide, do sliders/colorpickers for each component of the shadow. Which could then be exported from common/hud-customizer.ts.

Natanxp2:
EDIT: Already resolved

Cvars to remove

```
    mom_hud_map_name
    mom_hud_map_author
    mom_hud_map_stages
    mom_hud_map_type
    mom_hud_map_tier

    mom_hud_strafetrainer_mode
    mom_hud_strafetrainer_color_enable
    mom_hud_strafetrainer_dynamic_enable
    mom_hud_strafetrainer_flip_enable
    mom_hud_strafetrainer_buffer_size
    mom_hud_strafetrainer_min_speed
    mom_hud_strafetrainer_stat_mode

    mom_hud_stickybomb_charge
    mom_hud_stickybomb_charge_text

    mom_hud_stickybomb_count

    mom_hud_conc_timer
    mom_hud_conc_timer_countdown

    mom_hud_conc_entpanels_enable

    mom_hud_status

    mom_hud_comparisons

    mom_hud_timer

    mom_hud_jstats_enable
    mom_hud_jstats_takeoff_speed
    mom_hud_jstats_speed_delta
    mom_hud_jstats_takeoff_time
    mom_hud_jstats_time_delta
    mom_hud_jstats_strafe_count
    mom_hud_jstats_strafe_sync
    mom_hud_jstats_speed_gain
    mom_hud_jstats_yaw_ratio
    mom_hud_jstats_height_delta
    mom_hud_jstats_distance
    mom_hud_jstats_efficiency
    mom_hud_jstats_first_jump
    mom_hud_jstats_interval
    mom_hud_jstats_log_history

    mom_hud_spectator_list_show
    mom_hud_spectator_list_names
    mom_hud_spectator_list_max_name

    mom_hud_weapon_select

    mom_hud_df_groundboost_enable
    mom_hud_df_groundboost_overshoot
    mom_hud_df_groundboost_crash_highlight
    mom_hud_df_groundboost_text_mode
    mom_hud_df_groundboost_text_color_mode

    mom_hud_df_jump_enable
    mom_hud_df_jump_text
    mom_hud_df_jump_max_delay

    mom_hud_comparisons

    mom_hud_weapon_select
    mom_hud_weapon_select_keys
```

Events to remove
```
    OnStrafeTrainerModeChanged
    OnStrafeTrainerColorModeChanged
    OnStrafeTrainerDynamicModeChanged
    OnStrafeTrainerDirectionChanged
    OnStrafeTrainerBufferChanged
    OnStrafeTrainerMinSpeedChanged
    OnStrafeTrainerStatModeChanged
    OnStrafeTrainerStatColorModeChanged
    OnDefragHUDGroundboostChange
    MomentumSpecListMaxNamesUpdate
    OnJumpStatsCFGChange
    DFJumpMaxDelayChanged
```
    
+ `OnStickyPanelStateChanged` does not need prevState anymore

scripts/hud/customizer.ts
Natanxp2: It's extremely bad, locking up the game for up to 10s on whenever I click the dropdown and 1-2s lags whenever I scroll

scripts/hud/customizer.ts
Natanxp2: It's because CustomizerPropertyType.NONE has no value

EDIT: I meant to link to line 165 in customizer.ts, I have no idea how to do it on github, this ui sucks

scripts/hud/customizer.ts
Natanxp2: in order to use a flat map I would have to do this I think
```js
const childrenStyleIDs = new Set(
			Object.values(component.dynamicStyles)
				.map((style) => style.properties.children)
				.filter(Boolean)
				.flatMap((children) => (Array.isArray(children) ? children : [children]))
				.map((c) => c.styleID)
		);
```

Fun, but less readable than using a standard loop I think. I'll change it to object.values, no idea why I haven't used it here

tsa96: You could skip the filter with a ?? [] in previous map, but yeah sure imperative approach is probably better

scripts/hud/customizer.ts
tsa96: Mmh, tempted to expose OS version via MomentumVersionAPI in that case so we can enable conditionally, seems fine on Windows

scripts/hud/customizer.ts
tsa96: But the names of the enum members refer to what UI is generated for them, not anything about the value - maybe the enum itself is poorly named. Also if we were to add some other value-less element (idk labels explaining sections or something) it'd be simpler to add new elements here, rather than branching based on other properties of the DynamicStyle.

scripts/hud/customizer.ts
Natanxp2: I usually prefer keeping things in one line and get annoyed at prettier wrapping lines constantly :D

I don't mind using braces though if that's what you prefer, I'll keep that in mind for the future

scripts/common/hud-customizer.ts
Natanxp2: I thought it could be used as just a label or separator and didn't have a better idea for a name. It basically ended up only being used as a container. Don't know if we need to keep the ability to do this
image


If not I can rename that to container and probably also make sure `expandable` property is ignored on it and defaulted to true

scripts/hud/customizer.ts
Natanxp2: It is in there. This TODO also referred to Hud_Init event but I kept it in here cause we were still discussing what to use

scripts/hud/ammo.ts
Natanxp2: Replaced all of them with Gamemode Categories ( in every hud panel )

scripts/hud/customizer.ts
Natanxp2: I renamed --children to --has-children and added the first class to $.CreatePanel instead of making a separate call

--children-hidden needs to be a separate call though because it should only be added if parent is expandable 

( you made a comment on an outdated version of this file )

scripts/hud/customizer.ts
Natanxp2: Fixed

scripts/hud/strafe-trainer.ts
Natanxp2: You're right, I don't know why I didn't think of that. I'll replace all of them but first wanna figure out the text shadow stuff

scripts/hud/status.ts
Natanxp2: Text shadow with labels will definitely be way too wide, without labels maybe it will look good enough but then it won't be understandable to anyone.

In theory I could add separate styles for each part of the text shadow but I wouldn't be able to export it from common/hud-customizer, that would only work for mostly static panels.

When fonts change color dynamically ( like speedometer or keypress hud ) we would need to add 4 styles per color just for the text shadow.

I can add SHADOW_PICKER without any labels if you think that's good enough but I kinda thought it was "too much customization". 

Black already looks good and even scaled up doesn't look bad.
Adding all those options creates a lot of clutter in the UI for not a lot of benefit imo but maybe I'm wrong and people would really want to be able to customize the text shadow

scripts/hud/customizer.ts
Natanxp2: Sorry, I don't think I fully understand the issue here. I don't think we will be any more value-less elements in the future but even if we were they would all be handled the same, no? Just modifying the check on the line I mentioned would be enough.
If what you'd like to have a more concrete separation between elements with values and elements without values I can do that. I just thought this was the simplest way to handle what will probably be the only exception

scripts/hud/customizer.ts
Natanxp2: It's fixed now

scripts/hud/customizer.ts
Natanxp2: Would be great if you could figure out why this causes lag now that you're on linux, if that's not possible then the OS version is probably the best approach

scripts/hud/customizer.ts
Natanxp2: I went through all the comments and added spaces where appropriate

scripts/hud/customizer.ts
tsa96: Oh I'm really sorry, I was reviewing on this on my phone and thought this was addressing a different comment (which I definitely made but now can't find?). This comment was more of a joke, since `any` is equivalent to all other types, it's equivalent to the type `undefined`. And it's a bit of a confusing aspect of TypeScript, whether `?:` means a type is possibly `undefined` or whether it possibly that property doesn't exist on a object.

The thing I wanted to address was the name of `CustomizerPropertyType.NONE`, to suggest renaming it to something like `CustomizerPropertyType.CONTAINER`.

tsa96: Oh never mind, I found the comment, seems like that's resolved now anyway.

scripts/common/hud-customizer.ts
tsa96: Oh I see. That's basically what I addressing in my comment on the wrong thread here https://github.com/momentum-mod/panorama/pull/351#discussion_r3172547295 where I suggested it'd be better to just have CustomizerPropertyType.CONTAINER and others instead of branching off of additional properties as you're doing here - is `expandable` used for anything besides this? Isn't it possible to just tell if it's expandable by if it's CustomizerPropertyType.CONTAINER, whilst labels would get CustomizerPropertyType.LABEL? I may be misunderstanding.

scripts/hud/status.ts
tsa96: Doing an reusable expandable block of x offset/y offset/blur/color that gets exported from the common/ file seems good, collapsed by default but gives people the chance to adjust if they want. Could maybe add a referencedvalue thingy that refers to common shadow styles, again in an expandable block.

scripts/hud/customizer.ts
tsa96: Maaaaybe though it might be some time before I can compile on Linux anyway, swapping between both OSs atm and taking a while to configure Linux

layout/hud/ahop-sprint.xml
```diff
@@ -0,0 +1,17 @@
...
+	<styles>
+		<include src="file://{resources}/styles/main.scss" />
+	</styles>
+	
+	<scripts>
+		<include type="module" src="file://{scripts}/hud/ahop-sprint.ts" />
+	</scripts>
+
+	<MomHudAhopSprint>
+		<Panel id="AhopSprintCurveWrapper" class="h-fit-children">
```
tsa96: Just a comment since this isn't worth fixing - `fit-children` is the default value for `width` and `height` properties.

scripts/hud/ahop-sprint.ts
```diff
@@ -0,0 +1,340 @@
...
+					}
+				}
+			}
+		});
+	}
+
+	updateStyles() {
+		if (Config.type === 'curve') {
+			this.panels.curveWrapper.style.visibility = 'visible';
+			this.panels.dot.style.visibility = 'collapse';
```
tsa96: All panels have a bool `visible` property that's preferable and functionally equivalent to this

scripts/hud/chat.ts
```diff
@@ -0,0 +1,88 @@
...
+						const text = panel.FindChild('Text')!;
+						text.style.fontFamily = value;
+						// TODO: Need a better approach than this. Why does this work?
+						// If works because causes a repaint, why doesn't the repaint setting the font do it?
+						// Currently the size is always wrong when changing or for new messages, until they're
+						// mouseovered.
+						// Looks like setrepaint(full) *is* getting called in afterstylesapplied... so we need to do
+						// more than repaint to update styles? worth looking into comment on styles.cpp l952
+						// @ts-expect-error asdhjaskfodsf
+						panel.ApplyStyles(true);
```
tsa96: At this point I think we're just gonna go with it, we don't have time to dissect Panorama layouting. So we should add this to pano-typed and remove the TODO.

scripts/common/hud-customizer.ts
Natanxp2: as of right now this is the only way expandable is used but it doesn't have to be. For example we could have a margin setting that allows users to set all the margins. You could make that setting expandable to allow allow for modifying individual margins. This is how this would look
image

Essentially there are 2 systems for hiding children.
Expandable allows you to show/hide the entire container
showWhen property allows for showing/hiding individual styles based on parent value

You can mix them together. For example let's say you wanted to have the expandable style for margins from above but you only wanted to allow editing marginDown when the value of the parent is 2 ( nonsensical example, I know )
showWhen_expandable

Natanxp2: The thing that is missing from this system, that I completely forgot about cause it's not used anywhere is that I'm not handling properly cases like those above.
Currently if a parent is expandable and has a value, that value doesn't affect the children in any way.
I'm not sure what's the best way to handle that. Should the parent value change override children values?
That would work for the margin example but maybe are cases where that wouldn't be appropriate.
I was putting off deciding how to handle that until we have to actually use that option.

I'm currently leaning towards just keeping it as it is and letting individual components handle those cases in whatever way they need to be handled

scripts/hud/status.ts
Natanxp2: To better illustrate what I mean let's take keypress as an example
image


Those are styles for the labels under the keypress hud ( jump, duck, etc... )
Under every single color here there would need to be another expandable style with 4 options inside of it just for handling text shadow. That adds 5 visible labels, a total of 20 additional settings to handle through ts and makes the options less readable.

They also cannot be exported from common/ file, those colors are dynamically changed and they need to be added to the keypress config, every case needs to be handled individually. Same goes for speedometer and most other panels.

A reusable style could basically only be used for map-info and status panels unless I'm forgetting a panel.
Most panels need to store the values in variables for JS to apply, they can't just set styles on panels permanently

What I'm very concerned about is adding a bunch of customization options that most people don't need simply because we can. With the amount of options we have, even with all the grouping and organization I try to do it still sometimes quite a lot in a single menu.

I agree that approach I went with isn't the best, it's probably one thing I dislike the most about my changes but I think we should really be concerned about prioritizing readability of settings over adding all potential styling options.

As I said, it's possible that I'm wrong here and this is something important that should be added but maybe we should wait for community feedback on this. If people really need it  we can add it anytime, it's really simple to do and refactoring all the panels will take less than an hour

scripts/hud/strafe-trainer.ts
Natanxp2: Fixed this in the newest commit.
Is there a reason why rgbaTuple is 0-255 for all values including alpha? I didn't change it yet in case there is some reason for it but it's kind of annoying to work with

scripts/common/hud-customizer.ts
tsa96: Yeah, very case-by-case. Not sure you'd want parents updating the children for margins, or really much else for that matter. Your approach seems fine for now.

scripts/hud/customizer.ts
```diff
@@ -0,0 +1,2164 @@
...
+					Object.entries(this.dynamicStyles)
+						.filter(([_, dynamicStyle]) => dynamicStyle.value !== null)
+						.map(([styleID, dynamicStyle]) => [styleID, dynamicStyle.value])
+				)
+			})
+		};
+	}
+
+	/** Reset a component to its original state. */
+	// TODO: how tf will this work with groups lol
```
tsa96: Can remove this comment

scripts/hud/customizer.ts
```diff
@@ -0,0 +1,2164 @@
...
+			styleValue = Component.referencedValues?.[refKey] ?? undefined;
+		}
+
+		if (style.properties.styleProperty && styleValue !== undefined) {
+			const modifiedValue = style.properties.valueFn?.(styleValue) ?? styleValue;
+
+			if (modifiedValue !== undefined) {
+				for (const panel of targetPanels) {
+					(panel as any).style[style.properties.styleProperty] = modifiedValue;
+					// @ts-expect-error wadsasdasd
```
tsa96, line 401: Isn't this in the typedefs now? If not should add

scripts/hud/customizer.ts
```diff
@@ -0,0 +1,2164 @@
...
+		this.panels.settings.SetDialogVariable('preset_name', `${this.currentGamemodeInfo.name} Presets`);
+
+		const createDropdown = (parent: GenericPanel) => {
+			const panel = $.CreatePanel('Panel', parent, '');
+			panel.LoadLayoutSnippet('dynamic-dropdown');
+			panel.SetDialogVariable('name', 'Preset');
+			const dropdown = panel.FindChildTraverse<DropDown>('DropDown')!;
+
+			// Filters presetList + this.unsavedPresets to only those that match gamemodeID_<name> pattern, then strips gamemodeID + _
+			const gamemodeIdsByLength = [...GamemodeInfo.values()]
```
tsa96: Don't worry about changing this but just pointing out this can now be `.values().toArray()` since iterator methods were added a few years back

scripts/hud/customizer.ts
```diff
@@ -0,0 +1,2164 @@
...
+			if (isObject(accValue) && isObject(objValue)) {
+				acc[key] = deepMerge(accValue, objValue);
+			} else {
+				acc[key] = objValue;
+			}
+		});
+
+		return acc;
+	}, {} as any) as T;
+}
```
tsa96: Move these to `functions.ts`. In fact, we already have a deepMerge in there!

scripts/modals/popups/hud-customizer-layout-name.ts
```diff
@@ -0,0 +1,51 @@
...
+		if (callbackHandle !== -1) {
+			UiToolkitAPI.InvokeJSCallback(callbackHandle, this.panels.textEntry.text.toLowerCase());
+		}
+
+		UiToolkitAPI.CloseAllVisiblePopups();
+	}
+
+	validateTextEntry(): boolean {
+		const text = this.panels.textEntry.text.toLowerCase();
+		const regex = /^[\w-]+$/;
```
tsa96, line 36: Doesn't look like this allows underscore?

styles/main.css
tsa96, line 1: No

.prettierrc.json
```diff
@@ -1,6 +1,6 @@
 {
 	"printWidth": 120,
-	"tabWidth": 4,
+	"tabWidth":4,
```
tsa96: ? This might've been me

styles/config.scss
```diff
@@ -75,7 +75,7 @@ $light-200: rgba(255, 255, 255, 0.075) !default;
 $light-300: rgba(255, 255, 255, 0.1) !default;
 $light-400: rgba(255, 255, 255, 0.125) !default;
 $light-500: rgba(255, 255, 255, 0.15) !default;
-$light-600: rgba(255, 255, 255, 0.2) !default;
+$light-600: rgba(0, 0, 0, 0.2) !default;
```
tsa96: Both these colour changes look like mistakes on my part. #1896d3 is the official Momentum blue color, it's a major part of our branding, dunno why I changed it. pls revert!

scripts/modals/popups/hud-customizer-layout-name.ts
Natanxp2: \w is equivalent of [A-Za-z0-9_]

styles/main.css
Natanxp2: no?

.prettierrc.json
Natanxp2: I didn't touch this file

scripts/hud/customizer.ts
Natanxp2: I don't think I have access to types repo. I also forgot what it's supposed to be doing behind the scenes except for "fixing styling"

styles/main.css
tsa96: main.css/main.css.map don't need to be shipped. Maybe my change, but worth reverting

scripts/hud/ahop-sprint.ts
Natanxp2: fixed

scripts/hud/chat.ts
Natanxp2: It's already in pano-typed, just an old comment. Removed the todo

scripts/hud/customizer.ts
Natanxp2: removed

scripts/hud/customizer.ts
Natanxp2: Moved the compares to functions.ts and removed this deepMerge. Using mergeDeep from functions.ts

styles/config.scss
Natanxp2: reverted this, main.css and main.css.map

styles/main.css
Natanxp2: reverted them

scripts/hud/customizer.ts
tsa96: The repo is public, if you checkout with submods should be able to rebase

scripts/hud/customizer.ts
tsa96: I've exposed this to Pano, now in use

Natanxp2:
Localization file: [localization.json](https://github.com/user-attachments/files/28351058/localization.json)

Natanxp2:
Localization json: [localization.json](https://github.com/user-attachments/files/28390337/localization.json)

Natanxp2:
[localization.json](https://github.com/user-attachments/files/28878625/localization.json)


Added UI_YES and UI_NO to the list. They are only used in a yes/no popup afaik

Natanxp2:
[hud.zip](https://github.com/user-attachments/files/28965866/hud.zip)
Probably last hud update before release, needed for tab menu changes

scripts/modals/popups/import-export-settings.ts
```diff
@@ -132,6 +132,7 @@ class ImportExportSettingsHandler implements OnPanelLoad {
 	}
 
 	exportSelectAll() {
+		//@ts-expect-error Argument not typed
```
tsa96: What's happened here? `SetFocus` doesn't take any args. Not sure why/when this started being an error, but looks like `true` arg should be removed, rather than this comment being added.

scripts/types
tsa96, line 1: Our branch on the `types` submod needs to be rebased on its latest `main`, then make a PR with our changes. The stronger color types might need discussion, worth seeing it P2:CE still passes type checks with it.

scripts/common/hud-customizer.ts
```diff
@@ -0,0 +1,285 @@
+import { rgbaStringToTuple } from 'util/colors';
+// TODO: *VERY* detailed docs here, the types are unreadable!
```
tsa96: Docs comments are decent now, can remove this todo.

scripts/hud/customizer.ts
```diff
@@ -0,0 +1,2246 @@
...
+import * as LayoutUtil from 'common/layout';
+import * as Enum from 'util/enum';
+import { traverseChildren } from 'util/functions';
+import { GamemodeInfo, GamemodeInfoList } from 'common/gamemode';
+import { Gamemode } from 'common/web/enums/gamemode.enum';
+import { PanelHandler } from 'util/module-helpers';
+import { mergeDeep, compareDeepIgnoreNull, compareDeepAsymmetric } from 'util/functions';
+import { CopyToOtherPresetsData } from 'modals/popups/hud-customizer-copy-to-other-presets';
+
+// TODO: need to do a *ton* of localization when this is done, including components!
```
tsa96: This is all done in that localization.json file right? If so, remove todo

scripts/hud/customizer.ts
```diff
@@ -0,0 +1,2246 @@
...
+		$.RegisterForUnhandledEvent('HudCustomizer_Closed', () => {
+			this.disableEditing();
+			if (this.customizerReady) this.save();
+		});
+
+		// TODO: Below todo is *probably* fine now using events like this, but be very careful everything
+		// is getting registered okay.
+		// TODO (Old): I *think* we're gonna need this event to cover all cases where the HUD is reloaded.
+		// Looks like some stuff like HudSpecInfoHandler listening for PanelLoaded is getting
+		// called later than this though...
```
tsa96: If we've confirmed this is all working, can remove. Keen we get rid of all these leftover TODO of mine

scripts/hud/customizer.ts
```diff
@@ -0,0 +1,2246 @@
...
+
+		//Check if remaining components are the same as current preset
+		//This is done to not save unnecessarily when switching between presets
+		if (compareDeepIgnoreNull(saveData, savedPreset)) return;
+
+		const gamemodeID = this.gamemodeInfo.id;
+		const fullPresetName = `${gamemodeID}_${this.currentPreset}`;
+		// Serialization done in C++.
+		//TODO: Why the fuck is this saving with null values
+		const isSaved = this.panels.customizer.saveLayout(fullPresetName, saveData);
```
tsa96: Is this resolved?

scripts/hud/customizer.ts
tsa96: Still needs adding, are you really not able to access https://github.com/StrataSource/pano-typed ? Behind the scenes (I *think* this is accurate, should be accuare enough), it marks a panel as having had its styles changed, and that it needs a full repaint on next layout traversal.

.prettierrc.json
tsa96: Reverted it

tsa96:
Re-reviewed the core files, we're almost there

scripts/hud/customizer.ts
Natanxp2: It's not. I have no idea what's going on with it, doesn't really affect anything, it was just annoying me

Natanxp2:
I went through all todo comments and removed/updated them to the current state of things

scripts/hud/customizer.ts
Natanxp2: Since it doesn't affect the functionality at all I'm resolving it. I modified the comment to something more sensible

scripts/hud/comparisons.ts
```diff
@@ -114,7 +137,7 @@ class HudComparisonsHandler {
 			// If timer just finished and we're not watching a replay, don't update - otherwise
 			// comparison will get set to the run you just finished when you PB which we
 			// obviously don't want to happen.
-			if (!(state === Timer.TimerState.FINISHED && this.controlledReplayID === null)) {
+			if (state !== Timer.TimerState.FINISHED && this.controlledReplayID !== null) {
```
Panzerhandschuh, line 140: `if (!(state === Timer.TimerState.FINISHED && this.controlledReplayID === null)) {`
is not the same as
`if (state !== Timer.TimerState.FINISHED && this.controlledReplayID !== null) {`

Was this change intentional? If not, it should be:
`if (state !== Timer.TimerState.FINISHED || this.controlledReplayID !== null) {`

scripts/hud/sticky-count.ts
```diff
@@ -7,38 +11,188 @@ export enum StickyState {
...
-	[StickyState.NOSTICKY, 'sticky-panel--nosticky'],
-	[StickyState.ARMING, 'sticky-panel--sticky-arming'],
-	[StickyState.ARMED, 'sticky-panel--sticky-armed'],
-	[StickyState.BLOCKED, 'sticky-panel--sticky-blocked']
-]);
+// Values get overridden on load, assign defaults in /cfg/hud_default.kv3
+const StateColors: Record<StickyState, string> = {
+	[StickyState.NOSTICKY]: 'rgba(35, 50, 57, 1)',
+	[StickyState.ARMING]: 'rgba(35, 50, 57, 1)',
+	[StickyState.ARMED]: 'gradient(linear, 0% 0%, 100% 0%, from (rgba(24, 150, 211, 1), to(rgba(113, 240, 255, 1)))',
```
Panzerhandschuh, line 18: Is this a valid style? It looks like the parenthesis are unbalanced (5 opening parenthesis, ~~3~~ 4 closing). Similar issue on lines 19, 78, 86, 94, and 102. Also the 'from' and 'to' are inconsistent ('from' has a space after it, 'to' does not).

scripts/hud/replay-controls.ts
```diff
@@ -235,16 +313,22 @@ class ReplayControlsHandler {
 		this.sliderSegments.forEach(({ startTime, endTime, panel, progressPanel }) => {
 			// For iters on later segments after current one
 			if (foundActive) {
-				panel.SetHasClass('replaysegment--active', false);
+				if (this.seekBarColor) {
```
Panzerhandschuh, line 316: I noticed the replay seek progress bar was invisible in the map selector rework branch. Maybe seekBarColor and seekBarColorActive are undefined by default.

scripts/hud/sticky-count.ts
Panzerhandschuh: I think this still parses fine, but might be worth cleaning up.

scripts/hud/customizer.ts
```diff
@@ -0,0 +1,2238 @@
...
+			// Left Anchored: Distance from left edge is 1:1
+			return this.offsetX;
+		} else if (this.offsetX > (VIRTUAL_WIDTH / 3) * 2) {
+			// Right Anchored: Re-apply distance from the virtual right edge to 1920
+			return this.offsetX - VIRTUAL_WIDTH + DEFAULT_WIDTH;
+		} else {
+			// Center Anchored: Re-apply distance from the virtual center to 960
+			return this.offsetX - VIRTUAL_CENTER_X + DEFAULT_WIDTH / 2;
+		}
+	}
```
Panzerhandschuh, line 310: This can cause panels to drift between saves on non-19:6 aspect ratios (I haven't tested this but it was flagged by Opus).

```suggestion
    getDefaultOffsetX() {
		const anchor = this.width ? this.offsetX + this.width / 2 : this.offsetX;

		if (anchor < DEFAULT_WIDTH / 3) {
			// Left Anchored: Distance from left edge is 1:1
			return this.offsetX;
		} else if (anchor > VIRTUAL_WIDTH - DEFAULT_WIDTH / 3) {
			// Right Anchored: Re-apply distance from the virtual right edge to 1920
			return this.offsetX - VIRTUAL_WIDTH + DEFAULT_WIDTH;
		} else {
			// Center Anchored: Re-apply distance from the virtual center to 960
			return this.offsetX - VIRTUAL_CENTER_X + DEFAULT_WIDTH / 2;
		}
	}
```

Panzerhandschuh (approved):

tsa96:
https://github.com/StrataSource/pano-typed/pull/58 types submod PR here

I can't approve this since I made the PR, but I'm happy with stuff

